PR #13, opened 2011-08-30
Allow extra co-ordinate values to be stored

lukeredpath:
Allow extra co-ordinate values to be stored, to support the extra values allowed in the GeoJSON spec.

This patch is in support to Issue #2 for rgeo-geojson.


teeparham:
Closing as stale.

PR #46, opened 2013-01-13
Parse 'NAN' values in WKT formats

sempervictus:
This commit adds a hackish gsub! to string arguments passed to
parse_wkt methods in the factories. Some geometric objects can
develop 'NAN' values when adding a dimension and WKT parsing of
these value fails.


teeparham:
Closing as stale. If this is still an issue, please re-open with some tests to reproduce the problem.

Issue #32, opened 2012-07-20
SRID 3857 (instead of 3785)


ghazel:
http://alastaira.wordpress.com/2011/01/23/the-google-maps-bing-maps-spherical-mercator-projection/

rgeo's "simple mercator" seems to use a set of parameters that don't fully match 3857 or 3785:

```
PROJCS["Popular Visualisation CRS / Mercator",
  GEOGCS["Popular Visualisation CRS",
    DATUM["Popular_Visualisation_Datum",
      SPHEROID["Popular Visualisation Sphere",6378137,0,AUTHORITY["EPSG","7059"]],
      TOWGS84[0,0,0,0,0,0,0],
      AUTHORITY["EPSG","6055"]],
    PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],
    UNIT["degree",0.01745329251994328,AUTHORITY["EPSG","9122"]],
  AUTHORITY["EPSG","4055"]],
  UNIT["metre",1,AUTHORITY["EPSG","9001"]],
  PROJECTION["Mercator_1SP"],
  PARAMETER["central_meridian",0],
  PARAMETER["scale_factor",1],
  PARAMETER["false_easting",0],
  PARAMETER["false_northing",0],
  AUTHORITY["EPSG","3785"],
  AXIS["X",EAST],
  AXIS["Y",NORTH]]
```

In 4b972c3c53a9862520a09d7f9df98f0f9dd8f59f the SRID was changed from one to the other, but I'm not sure either are quite correct.

PR #61, opened 2013-12-20
moves from the deprecated EPSG 3785 to EPSG 3857

ebenoist:
This pull request solves the issue brought up in #32. EPSG 3785 has been deprecated in favor of the otherwise identical EPSG 3857.


averas:
Any idea when this release comes? :-)

tdgs:
what about existing data? can you give some hints on how to migrate them?

ebenoist:
@tdgs The data itself doesn't need to be migrated as EPSG 3785 is identical to EPSG 3857 in every way but the SRID so nothing needs to be reprojected. You will need to change the srid for the data you're using though.

Issue #60, opened 2013-11-25
rgeo does not recognize Ubuntu 13.10's GEOS


JasonBarnabe:
I installed libgeos-dev from Synaptic package manager. This installs to /usr/lib/libgeos-3.3.3.so, which rgeo apparently can't find. After running

```bash
sudo ln -s /usr/lib/libgeos-3.3.3.so /usr/lib/libgeos.so
```
and removing and reinstalling geos, things work.

Is this rgeo's problem, or is it GEOS's, or Ubuntu's, or mine? I don't know.

tneems:
I've also run into this on Ubuntu 14.04. The symlink fixed it for me too.

deivid-rodriguez:
This should be fixed now.

Issue #73, opened 2014-06-22
CAPIGeometryMethods not defined on Heroku


volkanunsal:
My app depends on being able to create a factory this way:

    factory = RGeo::Geos.factory(:srid => 4326)

This call returns a factory object on my machine, but it returns nil on Heroku. So I traced down the bug to [this line](https://github.com/rgeo/rgeo/blob/master/lib%2Frgeo%2Fgeos.rb#L75) where there is a check for whether `CAPIGeometryMethods` constant is defined. Oddly enough, it is not defined on Heroku, yet it's defined on my machine. And when it's undefined, the factory class does not get required into `RGeo::Geos`.

There is some circular logic here. The CAPIGeometryImpl is never defined because CAPI_SUPPORTED is false, and it is false because CAPIGeometryImpl is never required.

I'm not sure why that check is in place. Can anyone shed a light on it for me? How can I fix this?

Thanks.

PR #75, opened 2014-07-08
Use geos-config to compile/link correctly against GEOS.

nextdayflight:
This request changes extconf.rb to rely on geos-config to discover correct GEOS header/library paths *and* names, rather than just a list of paths.

This should fix:
https://github.com/rgeo/rgeo/issues/60
https://github.com/rgeo/rgeo/issues/73
and should re-instate the CAPI tests on Travis which are currently being skipped.


bbradbury:
@teeparham this PR looks good on travis minus one failing test on 1.8.7 -- RGeo GEOS support is currently broken on Ubuntu Trusty without it.

teeparham:
@bbradbury thank you for checking. We should drop support for 1.8.7, so that's not an issue.

deivid-rodriguez:
This can be closed since the changes have been merged in #120

PR #82, opened 2014-11-13
add #simplify for Polygon and MultiPoligon when using CAPI

pigoz:
Hello, these add support for for using simplify without passing from FFI.

I added this because Heroku is bugged and doesn't support FFI on the Cedar stack, and they have no clue how to fix it (see: https://github.com/ffi/ffi/issues/344).

I'm not sure why in the MultiPolygon test I needed to use `@lenient_factory` (normal factory would return `nil`), maybe you can help me there so that I can possibly change it.

Please let me know if I have to change something!

### DIFF
--- a/ext/geos_c_impl/geometry_collection.c
+++ b/ext/geos_c_impl/geometry_collection.c
@@ -479,7 +479,6 @@ static VALUE method_multi_polygon_centroid(VALUE self)
   return result;
 }
 
-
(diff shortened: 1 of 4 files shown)


pigoz:
Ping. Any interest in merging this? I think Travis is spitting out an unrelated error on the rbx server.

pigoz:
#95 is a more general version of this. So closing the PR in anticipation that gets merged.

PR #84, opened 2014-11-24
Fix Marshal and Psych support for native geos factories

cjab:
When a factory uses native geos for wkb/wkt generation/parsing Psych
and marshal functions were unable to serialize the generators/parsers.
It seems like the methods assumed that the generators/parsers were WKRep
objects. When generators and parsers are using geos they seem to return nil. This prevents encode_with and marshal_dump from calling _properties on nil.


njackson:
+1

PR #85, opened 2015-01-12
Incorrect calculation of proj_window.random_point

glampr:
Fixes the incorrect calculation of RGeo::Geographic::ProjectedWindow#random_point.
Specifically the x coordinate would usually be outside the bounds of the window.

### DIFF
--- a/lib/rgeo/geographic/projected_window.rb
+++ b/lib/rgeo/geographic/projected_window.rb
@@ -240,18 +240,14 @@ module RGeo
       end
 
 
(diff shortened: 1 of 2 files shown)


glampr:
Sorry, I had the correct method by mistake in another branch.

teeparham:
Thank you. Would you mind squashing those last 2 commits?

glampr:
Not sure I did it the right way, but I removed the wrong one!

PR #89, opened 2015-02-21
Fixed #88 where antimeridian flipped sign

### LINKED ISSUE #88: Creating a point at the north-pole antimeridian flips the longitude
While this is a degenerate case it is not uncommon

```ruby
> RGeo::Geographic.spherical_factory(srid: 4326).point(180,90)
=> #<RGeo::Geographic::SphericalPointImpl"POINT (-180.0 90.0)">
```

I would expect this to stay as `#<RGeo::Geographic::SphericalPointImpl "POINT (180.0 90.0)">`

ajturner:
Fixed issue where a point at `(180,0)` became `(-180,0)` due to an incorrect boolean inclusion test (i.e. `>= 180` should be `> 180`)

Added two test specs

Resolves #88


teeparham:
Thank you. Yes, both 180 and -180 are valid (and equal, in fact), so we should not change either. 

For future reference, please leave off the version change in PRs. Although in this case, I'll take it as a welcome hint that we need a new version. :+1:

ajturner:
Thanks!

And acknowledged on version numbers.

PR #102, opened 2015-07-01
Homebrew is not supported out of the box

AndreasHeiberg:
I just detailed how this was not working for me.

https://github.com/rgeo/activerecord-postgis-adapter/issues/190


deivid-rodriguez:
This should be working out of the box now.

PR #109, opened 2015-09-04
buffer_with_style method improvements

ynelin:


teeparham:
Please move the CAP_ROUND, CAP_FLAT, and CAP_SQUARE constants into a module:

```ruby
module EndCapStyle
  CAP_ROUND = 1
  # etc
end
```

Can we get the values for these constants from Geos, rather than setting them directly? 

It is weird that the EndCapStyle enum values do not match the Geos documentation - does anyone know why?

I found this:

http://geos.osgeo.org/doxygen/BufferParameters_8h_source.html

ynelin:
Need to try to do this >
```
Can we get the values for these constants from Geos, rather than setting them directly?
```

Anyway, this pr not relevant due https://github.com/rgeo/rgeo/pull/207

PR #112, opened 2015-09-30
add_check_validation_methods

ynelin:
Add mapping for 2 methods from geos:
- GEOSisValid_r => is_valid?
- GEOSisValidReason_r => is_valid_reason

### DIFF
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -988,6 +988,50 @@ static VALUE method_geometry_steal(VALUE self, VALUE orig)
   return self;
 }
 
+static VALUE method_geometry_is_valid(VALUE self)
+{
+  VALUE result;
+  RGeo_GeometryData* self_data;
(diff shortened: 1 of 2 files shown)


teeparham:
Thank you for this PR! Sorry it took so long for me to review it.

I see you used the method names from GEOS for consistency. I think we should rename the methods:

#### `valid?` instead of `is_valid?`

`valid?` is simpler & follows best practices for naming predicate methods (methods that return a boolean).

#### `invalid_reason` instead of `valid_reason`

This method returns the reason the polygon is invalid, so let's call it that. 

If it's valid, I assume the reason is nil. Can you add a test for that case?

ynelin:
Rename methods
When the polygon is valid it returns "Valid Geometry", I added test for this.

teeparham:
Something went wrong with the rebase, I think?

ynelin:
Yep. I fixed that

Issue #114, opened 2015-10-13
incompatible library version - proj4_c_impl.bundle (fatal)


eddietejeda:
I was getting the error `proj4_c_impl.bundle: [BUG] Segmentation fault at 0x00000000000440` but when I upgraded to ruby 2.2.3 and ran `gem pristine rgeo` as described in Issue #105  but when I ran the suggested fix,  I got the following error:

`~/.rvm/gems/ruby-2.2.3@civicinsightedge/gems/activesupport-4.2.4/lib/active_support/dependencies.rb:274:in 'require': incompatible library version - ~/.rvm/gems/ruby-2.2.3@civicinsightedge/gems/rgeo-0.3.20/lib/rgeo/coord_sys/proj4_c_impl.bundle (fatal)`

There was a StackOverflow post on this issue yesterday, which you can see [here](http://stackoverflow.com/questions/33089209/error-when-running-rails-app-incompatible-library-version-rgeo-proj4).

After some digging around, I figured it out what was wrong.

I opened up proj4_c_impl.bundle file in a hex editor and saw that the binary was being linked to  `/usr/local/lib/libruby.2.2.0.dylib`, which was strange because I was using RVM to manage my rubies. `libruby.2.2.0.dylib` was a symlink to `/usr/local/Cellar/ruby/2.2.3/lib/libruby.2.2.0.dylib` so that told me it was using a Ruby version installed with Homebrew. I am not sure why I installed the Homebrew version of Ruby, but I did.

I removed the Homebrew packges `ruby` and `ruby-build`, ran `gem pristine rgeo`, opened up `proj4_c_impl.bundle` and saw that was linking to `~/.rvm/rubies/ruby-2.2.3/lib/libruby.2.2.0.dylib` and the app loaded fine.

This leads to two other questions. First, do we want to support Homebrew's ruby? If so, we need to debug this further.

The second question is: Do we assume that Ruby Managers are common? I would say so, and whenever I use a Ruby manager I am assuming that the current Ruby environment takes precedent over any other Ruby versions I have installed.

So I prioritized the order of libraries, and gave the Ruby manager environment precedent, committed to a branch, tested within my application and it works fine. Here is the fix:

https://github.com/eddietejeda/rgeo/commit/975984b950f5ed9a3c1d726e8b06db700523828d

If there are no objections, I would like to make a pull request for this, but please feel free to discuss if there is a reason why we would not want to give precedence to Ruby Managers that I am not aware of.

[...]

PR #115, opened 2015-10-13
prioritize the lib and header directories of ruby managers

eddietejeda:


teeparham:
Original issue: #114

teeparham:
It seems like discovering the path like in #75 is a better solution than hard-coded paths.

PR #120, opened 2015-11-21
Passing tests on jruby

deivid-rodriguez:
Some changes I made while I was trying to run the tests in JRuby.


deivid-rodriguez:
I've cherry-pick some commits from #75, which is a better solution than my first approach. So this should also fix #60 and #73

tneems:
:+1:

teeparham:
Thanks for removing the ActiveRecord-related code in b0363cb . I have no idea what that is for, and it certainly does not belong in this gem.

I'll bump the version and release a new gem later today. This might call for a 1.0 release so people can more easily understand changes via SemVer.

Big thanks to @nextdayflight and @deivid-rodriguez !!! :trophy:

deivid-rodriguez:
Thanks @teeparham !

PR #123, opened 2015-11-23
Add RuboCop

### LINKED ISSUE #121: Comply with ruby style guide
There are a number of odd code conventions that we should correct. Of note:

####  Local variables ending with _

```ruby
def sum(x_, y_)
  x_ + y_
end
```

Better:

```ruby
def sum(x, y)
  x + y
end
```

#### Multiple blank lines 

```ruby
class X


  def y
    0
  end


end
```

Better:

```ruby
class X
  def y
    0
  end
end
```

Others:
* Unnecessary namespace resolution
* Long lines

In general, let's comply with https://github.com/bbatsov/ruby-style-guide

deivid-rodriguez:
Fixes #121


teeparham:
Awesome! All hail our new robot overlords.

Would you mind adding a rubocop config file & setting a preference for double quotes & re-running? I lot of the changes were changing double quotes to single, & I prefer all double quotes all the time.

deivid-rodriguez:
Of course! Hold on.

deivid-rodriguez:
Note that the majority of the code uses single quotes. Now I get ~ 16.000 offenses instead of ~ 13.000. Do you still want to do it?

teeparham:
"Yes, let's go with double quotes everywhere."

deivid-rodriguez:
"Ok" lol

deivid-rodriguez:
Done!

Note that one of the issues you mentioned in #121 is not solved (`RuboCop` don't seem to detect `_`-ended variables as offenses). But at least we are enforcing some code style now.

tneems:
What about https://houndci.com instead? It can be configured with rubocop preferences, but runs automatically on pull requests and it's free for open source projects.

deivid-rodriguez:
@tneems I like the idea behind `hound`, but when I've seen it used (activeadmin) I've found it way too noisy. They actually ended up disabling it in `activeadmin` for this reason. Maybe it's a matter of tweaking the configuration, I don't know.

@teeparham What do you think? Do you want me to enable hound and give it a try?

teeparham:
I'm not a huge fan of hound - I agree it seems too noisy.

I'm going to merge this soon after a little more review.

teeparham:
I made one minor change to the style rules (always use `raise` instead of `fail`) and re-ran `rubocop -a`. I rebased the autocorrect commit with the rule changes. So I'm closing this in favor of #124.

PR #124, opened 2015-12-03
Add Rubocop, correct style violations

### LINKED ISSUE #121: Comply with ruby style guide

teeparham:
Fixes #121 
Builds on #123 with a few rule changes


teeparham:
Note: the commits are shown in an incorrect order here. I rebased & amended a few commits. This is a bug with GitHub & the commits will be merged in the correct order & shown in the correct order in master after merging.

PR #127, opened 2015-12-22
Update README.md

gavit:
Added commands for RH based distro's like centos.
Also important to add the -dev packages!


deivid-rodriguez:
Can you add backquotes to the commands so they are prettier? :)

PR #136, opened 2016-01-14
[Addresses #74] Remove repeated consecutive points from line-strings

alexeiemam:
`BasicLinearRingMethods#_validate_geometry` currently ensures first and last points are matching
PR introduces a modification to ensure repeated consecutive coordinates are removed.
Updates PR #74 for `0.5.x` with more representative test.

### DIFF
--- a/test/common/polygon_tests.rb
+++ b/test/common/polygon_tests.rb
@@ -222,6 +222,39 @@ module RGeo
           polygon = @factory.polygon ring, [inner_ring]
           assert_equal(polygon.coordinates, coordinates)
         end
+
+        def test_ignores_consecutive_repeated_points
+          point1_ = @factory.point(0, 0)
+          point2_ = @factory.point(0, 0)
+          point3_ = @factory.point(0, 1)
+          point4_ = @factory.point(0, 1)
+          point5_ = @factory.point(0, 1)
+          point6_ = @factory.point(1, 0)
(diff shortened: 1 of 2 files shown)


test/common/polygon_tests.rb
```diff
@@ -222,6 +222,39 @@ def test_polygon_coordinates
...
+          poly1_ = @factory.polygon(exterior_)
+          assert(!poly1_.nil?)
+
+          line_string = poly1_.exterior_ring
+
+            case line_string.class.name
+            when "RGeo::Tests::GeosCAPI::TestPolygon"
+            when "RGeo::Geos::FFILinearRingImpl"
+            when "RGeo::Geos::CAPILinearRingImpl"
+            when "RGeo::Tests::GeosFFI::TestPolygon"
```
alexeiemam: This conditional was required in `0.3.20` but no longer appears to be relevant in `0.5.x`

Mr0grog:
This fix would be *extremely* helpful for me—I'm having to do the same thing in my own code right now because we've found that a lot of data being imported into a system I'm working on has consecutive duplicate points. Fixing this automatically when parsing would be really nice (and might save others like a good amount of time scratching their heads trying to understand what's wrong with their data or code).

teeparham:
This looks great. We're already modifying the points if the first & last are the same, so this seems like an excellent additional cleanup step to add. Thank you! :clap:

PR #138, opened 2016-02-14
Rakefile + test naming cleanup

tneems:
* Use [rake-compiler](https://github.com/rake-compiler/rake-compiler) for building c extensions
* Use built in minitest task as test runner
* Rename all test files from `tc_*.rb` -> `*_test.rb`
* Use built in RDoc task to build docs
* Remove trailing `_` from any remaining local variables


teeparham:
YESSSSSSSSSS

Issue #113, opened 2015-10-13
WKBParser w/spherical factory parses some geometries successfully, returns nil for others


wrkrb33:
Using RGeo 0.4.0.  The ewkb geometries I'm working with are polygons pulled from a postgis database.  I am using them in all sorts of contexts and know them to be valid.  However, parsing fails on some percentage of them (10-20%?).  I haven't been able to identify any common characteristic among the polygons that fail to parse.

```ruby
> factory = RGeo::Geographic.spherical_factory(:srid=>4326)
=> #<RGeo::Geographic::Factory:0x00000002a2c1a8.....
> parser = RGeo::WKRep::WKBParser.new(factory, :support_ewkb=>true)
 => #<RGeo::WKRep::WKBParser:0x000000029f0180.....
> ewkb1 = "0103000020E6100000010000000F000000216F17D9E40958C05A5095CF557D4640C4DF3FD9E40958C0AB65939B557D4640822F2F10F50958C0FC33FDDE557D4640C43D3EB6F60958C064A48FE0557D4640A71EF2C3480A58C0C2714B2F567D4640AD35C4779C0A58C0CBDC8F7F567D46401BF1CD2B9B0A58C00A2E324ACF7D4640A62DCDDF990A58C04C64D114487E4640BE647BB85B0A58C097D32477477E46400B6D73C8450A58C0192E363F477E464090325922260A58C091E54EEE467E4640E57311B1F10958C01BE6BD67467E4640520EAE60F30958C048AA5E23CE7D46402369D379E40958C0F3BB7579CE7D4640216F17D9E40958C05A5095CF557D4640"
[...]

PR #144, opened 2016-06-15
Remove duplicate points when casting rings+lines

sosso:
I ran into a similar issue like #113, so I made this fix.  If you'd like it behind a config parameter or other changes, feel free to let me know.

Here's a polygon that chokes RGeo:  `'MULTIPOLYGON(((-122.10033416748 37.6906110800046,-122.10033416748 37.6906110800046,-122.075443267822 37.6911544368997,-122.075443267822 37.6911544368997,-122.057418823242 37.6933278246668,-122.057418823242 37.6933278246668,-122.045230865479 37.6956369793712,-122.045230865479 37.6956369793712,-122.046947479248 37.6899334219492,-122.033815383911 37.6705735585369,-122.015447616577 37.675329080766,-122.034072875977 37.7018849197626,-122.034072875977 37.7018849197626,-122.00626373291 37.7126153927168,-122.045917510986 37.7423479509276,-122.045917510986 37.7423479509276,-122.111663818359 37.7352888797979,-122.111663818359 37.7352888797979,-122.120418548584 37.7071816741417,-122.120418548584 37.7071816741417,-122.107887268066 37.6997117828282,-122.107887268066 37.6997117828282,-122.10033416748 37.6906110800046)))'`

### DIFF
--- a/lib/rgeo/feature/types.rb
+++ b/lib/rgeo/feature/types.rb
@@ -246,11 +246,11 @@ module RGeo
             end
           elsif ntype_ == LinearRing
             if type_ == LineString
-              nfactory_.linear_ring(obj_.points.map { |p_| cast(p_, nfactory_, opts_) })
+              nfactory_.linear_ring(obj_.points.chunk{|x| x}.map(&:first).map { |p_| cast(p_, nfactory_, opts_) })
             end
           elsif ntype_ == LineString
             if type_ == Line || type_ == LinearRing
-              nfactory_.line_string(obj_.points.map { |p_| cast(p_, nfactory_, opts_) })
+              nfactory_.line_string(obj_.points.chunk{|x| x}.map(&:first).map { |p_| cast(p_, nfactory_, opts_) })
             end
           elsif ntype_ == MultiPoint
             if type_ == Point


teeparham:
Thank you. This seems reasonable. Can you add a test with the polygon you listed? (`types_test.rb` seems like a good place for it)

sosso:
Sure! The only thing is that it could be considered unexpected behavior,
given that it's altering the data it's parsing. Is just altering the data
in this manner in the spirit of the project? Or should it throw a warning
or error, be configurable, etc? I do think it's pretty harmless, but I
would personally like to know if my shape data had duplicate points, as
that problem chokes more GIS tools than just RGeo, and this fix would mask
that.

teeparham:
Good point - perhaps `#cast` is not the right place to do this. The `#cast` method is doing too much as it is. Perhaps we need a `#clean` method on `Polygon`.

sosso:
Perhaps, but that wouldn't cover this case. I suppose we could put it on
the factory, which the ActiveRecord adapter supports on a per-column basis?
Something similar to lenient multipolygon assertions.

Ume1931:
When this commit will be merged to master?

BuonOmo:
@sosso Thanks for the PR!

I cannot reproduce the issue with both ruby and capi factories with the current version of rgeo, hence I'll close the pull request, if one still encounters a related problem, please open a new issue

PR #147, opened 2016-06-29
Rename variables to p_ to prevent undefined variable error

### LINKED ISSUE #146: undefined local variable or method `p_' for RGeo::Feature:Module
I'm receiving `undefined local variable or method 'p_' for RGeo::Feature:Module` when casting from a GeometryCollection to a MultiPolygon like this:

```
RGeo::Feature.cast(geom_collection, RGeo::Feature::MultiPolygon)
```

Dschee:
Attempts to fix #146.

### DIFF
--- a/lib/rgeo/feature/types.rb
+++ b/lib/rgeo/feature/types.rb
@@ -256,23 +256,23 @@ module RGeo
             if type_ == Point
               nfactory_.multi_point([cast(obj_, nfactory_, opts_)])
             elsif type_ == GeometryCollection
-              nfactory_.multi_point(obj_.map { |_g_| cast(p_, nfactory_, opts_) })
+              nfactory_.multi_point(obj_.map { |p_| cast(p_, nfactory_, opts_) })
             end
           elsif ntype_ == MultiLineString
             if type_ == Line || type_ == LinearRing || type_ == LineString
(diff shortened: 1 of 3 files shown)


teeparham:
Thank you. This looks good. Would you mind adding some tests for the conversions?

Dschee:
I'm sorry, it's not clear to me where to begin to add tests for casting. I'm not really into this project and merely fixed this issue because the compiler / interpreter gave me clear instructions on what was wrong. I just fixed a variable naming problem without really knowing what's happening here.

So, I might not be the right guy to add tests for this except when there were already tests testing all the other cases then I could figure out tests for these cases, too. But I couldn't find any other usage of the `cast` method except for one within `proj4_test.rb` and that's not really a good example for testing this method, is it?

Dschee:
@teeparham: Any further issues in this MR? If not, it would be great if this could be merged as I'm currently having problems with the GEOS paths when including this gem using my fork instead of this original one. ^^

Dschee:
Since there is no reaction and I'm stuck with this issue I'm gonna create a new gem with my fork until this is merged ...

dmitry:
@teeparham ping

teeparham:
Fixed with #162

PR #154, opened 2016-10-23
node geometry collection

alibby:
A project I'm working on requires the intersections in a MultiLineString be noded for processing.
This functionality from the GEOS C interface was not exposed.  This PR allows for the a call
to #node on geometry collection types resulting in a noded version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
+dist: trusty
 sudo: false
 addons:
   apt:
     packages:
-    - libgeos-3.4.0
+    - libgeos-3.4.2
     - libproj-dev
 language: ruby
 rvm:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 addons:
   apt:
     packages:
-    - libgeos-3.3.8
+    - libgeos-3.4.0
     - libproj-dev
 language: ruby
 rvm:
--- a/ext/geos_c_impl/geometry_collection.c
+++ b/ext/geos_c_impl/geometry_collection.c
@@ -363,6 +363,23 @@ static VALUE method_multi_line_string_hash(VALUE self)
   return LONG2FIX(rb_hash_end(hash));
 }
 
+static VALUE method_geometry_collection_node(VALUE self)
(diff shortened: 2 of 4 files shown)


alibby:
When working on another PR I noticed that I should be making changes to the interface(s) for 
this stuff.  I'll try to get something pushed here in the next day or so.

alibby:
Curios if I could get someone to look at and potentially merge this?  Thanks in advance.

teeparham:
This looks good to me. Can you add a test to `rgeo/test/geos_capi/geometry_collection_test.rb`? (see rgeo/test/geos_capi/misc_test.rb for examples)

alibby:
@teeparham - Pushed a test, let me know if you feel more is needed.

alibby:
So it appears the geos function GEOSNode_r is not present until 3.3.4.

teeparham:
Try rebasing on latest master - I updated travis to use geos 3.3.8.

alibby:
So it appears we might be on a super old Ubuntu version on travis.  It appears to be 12.04, which is pretty old and was end of lifed in April of this year. (http://releases.ubuntu.com/12.04/).  What do you think about updating to 14.04, or even 16.04, and then I'll play with updating the geos version to something more current, say something in the 3.5 or 3.6 range.

alibby:
It appears GEOSNode_r was introduced in 3.4.0, which was released after Ubuntu 12.04.  I think updating the travis runtime is the way to go here.

teeparham:
It looks easy to upgrade to Trusty (14.04):

```
dist: trusty
```

https://docs.travis-ci.com/user/reference/trusty/

I'm ok to use 14.04 or 16.04. Also I'm ok with testing against the latest GEOS. The default version with 14.04 + apt is 3.4.2:

https://packages.ubuntu.com/trusty/libgeos-3.4.2

https://packages.ubuntu.com/search?keywords=libgeos&searchon=names

Please try updating the distribution + GEOS version in this PR, & I'm happy to use the newest versions we can get working.

alibby:
I like the looks of that!

PR #156, opened 2016-10-27
Linear project and interpolate

alibby:


alibby:
I think I've got this one cleaned up, mind looking at it @teeparham?  TIA.

teeparham:
Thanks! 👍

Issue #47, opened 2013-02-03
Building Polygon yields a nil result


vidriloco:
I attempted to build a simple polygon with geographic data with the following code: 

``` ruby
factory = RGeo::Geographic.spherical_factory(:srid => 4326)
# This satisfies a linear ring requirement: to be a closed linear string
a=[factory.point(-92.272935049926986,15.831411705496032)]
a << factory.point(-92.273079163754431,15.830847348631419)
a << factory.point(-92.272935049926986,15.831411705496032)

# Then when trying to build a linear ring I get nil. Am I missing an additional requirement?
linear_ring = factory.linear_ring(a)
>> nil
# Then tried building a line string 
linestring = factory.line_string(a)
>> Worked as expected
# Attempted to build a polygon, failed
factory.polygon(linestring)
>> nil
´´´

Also, from the extended documentation on: http://blog.daniel-azuma.com/archives/88
I found out that executing the code for creating a cartesian polygon as in the example also yields a nil object. 

Is this a bug or am I missing something? Thanks
```

babasbot:
@tilo is there a way to sort those points?

keithdoggett:
Closing this since the spherical factory now raises an error when an invalid linear ring is being initialized.

PR #183, opened 2018-03-03
Remove exception catching

teeparham:
When parsing objects, let them raise an exception rather than silently returning nil.

#158 #47


teeparham:
We already raise `RGeo::Error::InvalidGeometry` in several situations. This makes the behavior consistent.

To preserve the old behavior, add your own `rescue nil` wrapper. Here's a generic method:

```ruby
def safe_parse
  yield
rescue RGeo::Error::InvalidGeometry
  nil
end

geometry = safe_parse { factory.polygon(points) }
```

PR #203, opened 2019-05-06
Input validation / conversion for libgeos WKB parser

chrisjwu:
### Summary

Not sure if this is a bug or feature enhancement, but the libgeos implementation does not support hexidecimal string input; while the Ruby implementation does (see added test in test/geos_capi/misc_test.rb).

This becomes an issue because most WKB representations use hexidecimal strings rather than binary strings; such as in ActiveRecord::PostGIS.  As such we cannot always rely on receiving a binary encoded input string and need to handle both hexidecimal representations as well as binary (which again the Ruby Parser: RGeo::WKRep::WKBParser already does).

### Other Information

This doesn't solve the open tickets regarding performance issues; but it does provide a workaround for people willing to switch their code to use the WKB parser.

Since this is a string encoding issue between hex representations and binary; this does not effect WKT parsing.

On a related topic: ActiveRecord::PostGIS is currently hardcoded to use the ruby parsers rather than accept parser options.


chrisjwu:
On a related note: after playing around with this fix I noticed that RGeo::WKRep::WKBParser (and WKTParser) ignore the wkt_parser / wkb_parser options.  This is because those parse methods never call parse_wkt / parse_wkb when parsing and instead call the token_parsing methods directly (eg: point / line_string / ect).

This has the unfortunate side-effect that you can only use the Ruby parser in ActiveRecord::PostGIS even if you set the factory_generator options to set the wkt/wkb parser to :geos.

teeparham:
Thank you. Would you mind removing the change to the version number from this PR? I will do that when a new version is released.

chrisjwu:
Done, sorry I was a bit over eager on the version.

Also an update on my comment about WKRep::WKBParser.  After some more investigation; WKRep *is* passing each token through to the C factory (when parser is set to geos).  I was confused about that since benchmarking different parser options for WKRep has almost no impact.  Implies most of the time is being spent scanning the string rather than consuming tokens.

lib/rgeo/geos/capi_factory.rb
```diff
@@ -297,7 +297,11 @@ def parse_wkb(str_)
         if (wkb_parser_ = _wkb_parser)
           wkb_parser_.parse(str_)
         else
-          _parse_wkb_impl(str_)
+          if str_[0] =~ /[0-oa-fA-F]/
```
aleksejleonov: I believe, there should be `/[0-9a-fA-F]/` instead of `/[0-oa-fA-F]/`.

aleksejleonov: @chrisjwu

Also, WKB binary string always starts with `\x00` or `\x01`.
> The first byte indicates the byte order. 00 for big endian, or 01 for little endian.
> https://mariadb.com/kb/en/library/well-known-binary-wkb-format/

So, it would be more efficient to detect binary WKB string this way:
```ruby
if str_[0] == '\x00' || str_[0] == '\x01' 
```

Similar way of detection used in https://github.com/rgeo/activerecord-postgis-adapter/blob/744bed7e276135be83ebfb6c498273766f918850/lib/active_record/connection_adapters/postgis/oid/spatial.rb#L99.

aleksejleonov: Fixed in #208.

PR #205, opened 2019-07-07
Add geometry point_on_surface method

andrejreznik:
### Summary

This PR implements the `Geometry#point_on_surface` method using GEOS CAPI. There is an implementation of `MultiPolygon#point_on_surface` already in the master branch. So, this PR just move existed logic to base feature.

### Other Information

Other features like point and lines, not only polygons should have `point_on_surface` method. It would be more consistent.  In PostGIS [ST_PointOnSurface](https://postgis.net/docs/ST_PointOnSurface.html) works fine with all feature types.

### DIFF
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -1053,6 +1053,21 @@ static VALUE method_geometry_invalid_reason(VALUE self)
   return result;
 }
 
+static VALUE method_geometry_point_on_surface(VALUE self)
+{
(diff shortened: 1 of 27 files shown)


ynelin:
@teeparham maybe you can merge this or @andrejreznik need to improve something?

teeparham:
Thanks! 

It would be nice to add/fix the tests with floating point issues, but that can be done in a future PR.

teeparham:
Released in 2.1.0.

PR #207, opened 2019-08-07
Fix buffer with style params types for Geos

ynelin:
### Summary
There was a mistake with params types for GEOS BufferWithStyle_r


teeparham:
Thanks! 👍

teeparham:
Released in 2.1.0.

PR #209, opened 2019-09-03
Add centroid calculation for spherical polygons

### LINKED ISSUE #142: Support centroid for SphericalPolygon
When using centroid on a SphericalPolygon I got the error:
RGeo::Error::UnsupportedOperation: Method Surface#centroid not defined.

Any reason centroid is not supported for SphericalPolygon?

allknowingfrog:
### Summary

Implement `centroid` for SphericalPolygon. Resolves #142.

### Other Information

This logic is borrowed directly from [geokit](https://github.com/geokit/geokit) with modifications to accommodate rgeo data structures. Note that the calculation assumes Euclidian geometry, which will become increasingly inaccurate as the size of the polygon grows. I think, however, that this will be close enough for most uses. The math for a perfect calculation seems to be vastly more complicated. See [this](https://stackoverflow.com/questions/19897187/locating-the-centroid-center-of-mass-of-spherical-polygons) for example.

### DIFF
--- a/lib/rgeo/geographic/spherical_feature_methods.rb
+++ b/lib/rgeo/geographic/spherical_feature_methods.rb
@@ -135,23 +135,15 @@ module RGeo
       def centroid
         return super unless num_interior_rings == 0
 
-        points = coordinates.first
-
         centroid_lat = 0.0
         centroid_lng = 0.0
         signed_area = 0.0
 
-        # Iterate over each element in the list but the last item as it's
-        # calculated by the i+1 logic
-        points[0...-1].each_index do |i|
-          x0 = points[i][0]
-          y0 = points[i][1]
-          x1 = points[i + 1][0]
-          y1 = points[i + 1][1]
-          a = (x0 * y1) - (x1 * y0)
-          signed_area += a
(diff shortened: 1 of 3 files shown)


lib/rgeo/geographic/spherical_feature_methods.rb
```diff
@@ -130,5 +130,36 @@ def length
...
+
+        points = coordinates.first
+
+        centroid_lat = 0.0
+        centroid_lng = 0.0
+        signed_area = 0.0
+
+        # Iterate over each element in the list but the last item as it's
+        # calculated by the i+1 logic
+        points[0...-1].each_index do |i|
```
teeparham: We could use `#each_cons` to iterate, and use `#exterior_ring` to get point objects instead of coordinates. Something like this should work:

```ruby
exterior_ring.points.each_cons(2) do |p0, p1|
  area = (p0.x * p1.y) - (p1.x & p2.y)
  # ...  
```

allknowingfrog: Yeah, that is much nicer.

teeparham:
That looks great. Thank you!

PR #210, opened 2019-09-17
Typo in boundary

mitchless:
### Summary

When updating our gems after upgrading Ruby to 2.6, I noticed our specs failing in `lib/rgeo/impl_helper/basic_geometry_collection_methods.rb` for rgeo 2.1.1. The failure traced back to a change related to removing underscores from the impl_helper directory.

I did not find any issue related to this bug.

This pull request adds a test to reproduce the failure we noticed in our code as well as a fix to the affected method.


teeparham:
Thanks!

PR #213, opened 2019-10-30
Change SphericalMath::PointXYZ to implement BigDecimal instead of Float

### LINKED ISSUE #212: Some "valid" polygons are treated as invalid by rgeo
Some polygons that are valid with MySQL or QGIS are causing error in rgeo.

### Steps to reproduce

The following code results in exception.

RGeo::Geographic.spherical_factory.parse_wkt("POLYGON ((126.968027438107 37.5816708669373, 126.96804356077827 37.58168793850955, 126.96804348189114 37.5816870915542, 126.96804311029837 37.58168238826706, 126.96803990177084 37.581642824769546, 126.968027438107 37.5816708669373))")

### Expected behavior

It should be parsed without problem.

### Actual behavior

It throws the following exception.

RGeo::Error::InvalidGeometry (LinearRing failed ring test)

### System configuration

**Ruby version**: 2.5.0p0

**OS**: Mac OS X 10.14.6

keithdoggett:
### Summary

Resolves #212 

Valid polygons will sometimes fail the linear ring test in spherical factories if they have a high point density. The test erroneously detects that arcs of the polygon intersect causing it to fail the `is_simple?` condition. After some testing, I found out the issue has to do with floats lacking enough precision in the spherical coordinate system.

To fix this, I cast the floats to `BigDecimal`s before any calculations are done on them. This has so far worked for all of the cases I've found of the issue.

### Other Information

I attempted to integrate the `BigMath` library into the class to handle trigonometric and sqrt calculations but it made the tests take about 240x longer. I didn't notice any increase in accuracy by using it so removed it.

### DIFF
--- a/lib/rgeo/geographic/factory.rb
+++ b/lib/rgeo/geographic/factory.rb
@@ -43,6 +43,7 @@ module RGeo
           @coord_sys = CoordSys::CS.create_from_wkt(@coord_sys)
         end
         @lenient_assertions = opts[:uses_lenient_assertions] ? true : false
+        @uses_decimals = opts[:uses_decimals] ? true : false
         @buffer_resolution = opts[:buffer_resolution].to_i
         @buffer_resolution = 1 if @buffer_resolution < 1
 
@@ -107,6 +108,7 @@ module RGeo
           "wktp" => @wkt_parser.properties,
           "wkbp" => @wkb_parser.properties,
           "lena" => @lenient_assertions,
+          "decm" => @uses_decimals,
           "bufr" => @buffer_resolution
         }
         hash_["proj4"] = @proj4.marshal_dump if @proj4
@@ -139,6 +141,7 @@ module RGeo
           wkt_parser: symbolize_hash(data_["wktp"]),
           wkb_parser: symbolize_hash(data_["wkbp"]),
           uses_lenient_assertions: data_["lena"],
+          uses_decimals: data_["decm"],
           buffer_resolution: data_["bufr"],
           proj4: proj4,
           coord_sys: coord_sys
@@ -165,6 +168,7 @@ module RGeo
         coder["wkt_parser"] = @wkt_parser.properties
         coder["wkb_parser"] = @wkb_parser.properties
         coder["lenient_assertions"] = @lenient_assertions
+        coder["uses_decimals"] = @uses_decimals
         coder["buffer_resolution"] = @buffer_resolution
         if @proj4
           str = @proj4.original_str || @proj4.canonical_str
@@ -202,6 +206,7 @@ module RGeo
           wkt_parser: symbolize_hash(coder["wkt_parser"]),
           wkb_parser: symbolize_hash(coder["wkb_parser"]),
           uses_lenient_assertions: coder["lenient_assertions"],
+          uses_decimals: coder["uses_decimals"],
           buffer_resolution: coder["buffer_resolution"],
           proj4: proj4,
           coord_sys: coord_sys
@@ -295,6 +300,8 @@ module RGeo
           @support_m
         when :uses_lenient_assertions
           @lenient_assertions
+        when :uses_decimals
+          @uses_decimals
         when :buffer_resolution
           @buffer_resolution
         when :is_geographic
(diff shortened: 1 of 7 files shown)


jasonhr13:
Would love if this could get merged!

Kdoggett887:
If you need to use the fix immediately and you're using a Gemfile/Bundler, you can add this to the Gemfile to build from the branch.

`gem 'rgeo', git: "git://github.com/Kdoggett887/rgeo.git", branch: "fix-212"`

Just make sure to change it back once it's merged in.

If you use the branch and are still getting errors, post in the original issue and I can take a look.

Kdoggett887:
Instead of forcing `BigDecimal` to be used, `uses_decimals` can be passed into `spherical_factory` as an option. If it's used, `PointXYZ` will use decimals as its data type instead of floats. 

Latitudes and longitudes on points will still be represented as floats.

usage:
`factory = RGeo::Geographic.spherical_factory(uses_decimals: true)`

jdalt:
@keithdoggett I've been bitten by this issue several times, and I really appreciate your efforts to fix the issue. I was wondering what the performance implications of this change are. I created a test to validate that this patch fixes my issue and noticed the test taking another 1.3s (0.3s vs 1.3s).

I was also wondering whether there might be an alternative approach where we continue to use floats but only trigger `#intersects_arc?` when the float passes a certain configurable threshold.

keithdoggett:
@jdalt thanks for the reply and testing out the code. Definitely appreciate the feedback.

You're right that the performance with decimals instead of floats is not nearly as good. I tried to alleviate that somewhat with my last update to make the option to use decimals configurable at the factory level, but it's still slower in cases you need to use it and the way that option is passed around isn't the cleanest. 

Your approach sounds interesting, but I need a little bit of clarification. When you say the float passes a certain configurable threshold, do you mean distance from another point or how precise a coordinate the float is representing?

jdalt:
@keithdoggett My idea was to use a threshold for detecting arc intersection. I'm not certain this would be a viable solution though since it sacrifices correctness, and I'm realizing these rounding areas actually let several actually invalid polygons into my dataset.

Ultimately I've worked around this by passing `uses_lenient_assertions: true` to my spherical factory. I'm using a different factory to validate whether the geometry is simple and valid. I only need the spherical coordinates for certain area calculations. The performance cost was too much for my app, unfortunately. I only tested with ruby 2.5, perhaps ruby 2.6 performance will be better.

keithdoggett:
Gotcha. At this point, it does seem to be a trade-off between correctness and performance. 

One theory that I haven't gotten around to testing is that the inaccuracies arise from converting from a spherical system to cartesian in `PointXYZ`. There's a comment in the class that indicates that this was done to improve calculation speed and increase stability on some calculations, but this may be inadvertently creating validation errors.

I think that the solution you posed is the best at this point, though. Do the validations in a projection or some other factory and then spherical calculations in a factory with lenient assertions.

jdalt:
I'd be interested to see dummy examples of the following:
1. the projected coordinate system catches a false intersection that doesn't really exist in spherical coordinates
2. the projected coordinate system doesn't catch an intersection that only exists when calculated in spherical coordinates

I would imagine that one of these happens when the geometry is very small (the dense packing problem) and one happens when things are very large. It would be nice to develop some guidance as to when these are important.

For both of these edge case situations, the spherical factory with decimals would be necessary. For everything else you might as well use a CAPI backed projected coordinate system. Knowing the heuristic for which factory to select is key.

waissbluth:
hey there, did anybody find another workaround for this?

keithdoggett:
@waissbluth we haven't made any progress on this issue specifically, although there have been discussions about potential solutions that we'll try to implement at some point in the future.

But we have been reworking the validity handling for version 3 (#271) so that errors won't be raised on instantiation and validity will only be checked on an as-needed basis (ex. `poly.area`). This wouldn't fix the issue of incorrectly identifying invalid polygons, but it would reduce some of the friction with the current method of aggressively raising errors.

There is still the option to pass the `uses_lenient_assertions` option into the spherical factory and it should bypass the check. Is there a reason why that wouldn't work for your use case?

PR #219, opened 2020-05-10
Fix typo

taki:
### Summary

I fix typo.

### Other Information

none


taki:
I will close because it is inactive

PR #221, opened 2020-06-15
Added method offset_curve

tyfoan:
Added offset_curve method

### DIFF
--- a/ext/geos_c_impl/line_string.c
+++ b/ext/geos_c_impl/line_string.c
@@ -123,7 +123,29 @@ static VALUE method_line_string_coordinates(VALUE self)
   return result;
 }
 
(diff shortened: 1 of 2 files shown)


keithdoggett:
Thanks for submitting this, the PR looks good!

The only thing that I think should change is that the method should be moved into the `geos_line_string_methods`/`ProjectedLineStringMethods` module since the comment in the `geos_c.h` file indicates that the function only accepts linestrings.

From `geos_c.h`:

```c
/*
 * Only LINESTRINGs are accepted.
 * @param width : offset distance.
 *                negative for right side offset.
 *                positive for left side offset.
 * @return NULL on exception
 */
```

tyfoan:
@keithdoggett Could you review last commit please?

Issue #211, opened 2019-10-03
Frozen string when force_encoding in ffi_feature_methods


ldonnet:
When we call as_text method for RGeo::Geos::FFIFactory, an error raise about Frozen string.
That happen when we read geometry from database or create point in console. 
The problem seems too big that no one else find it, so I describe the way to reproduce it. 

### Steps to reproduce

```sh
2.6.1 :013 >   WGS84_PROJ4 = '+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs'
 => "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs" 
2.6.1 :014 > 
2.6.1 :015 > f = RGeo::Geos::FFIFactory.new(
2.6.1 :016 >           srid: 4326,
2.6.1 :017 >           proj4: WGS84_PROJ4,
2.6.1 :018 >           :wkt_parser => {:support_ewkt => true, :default_srid => 4326},
2.6.1 :019 >           :wkt_generator => {:type_format => :ewkt, :emit_ewkt_srid => true},
2.6.1 :020 >           :wkb_parser => {:support_ewkb => true, :default_srid => 4326},
2.6.1 :021 >           :wkb_generator => {:type_format => :ewkb, :emit_ewkb_srid => true}
2.6.1 :022?>       )
 => #<RGeo::Geos::FFIFactory:0x2b29f4103f58 srid=4326> 
2.6.1 :023 > f.point(2,3)aceback (most recent call last):
       16: from /home/luc/.rvm/gems/ruby-2.6.1/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:259:in `load_dependency'
       15: from /home/luc/.rvm/gems/ruby-2.6.1/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:287:in `block in load'
       14: from /home/luc/.rvm/gems/ruby-2.6.1/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:287:in `load'
       13: from /home/luc/projects/referentials/bin/rails:9:in `<top (required)>'
       12: from /home/luc/.rvm/gems/ruby-2.6.1/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:293:in `require'
       11: from /home/luc/.rvm/gems/ruby-2.6.1/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:259:in `load_dependency'
[...]

PR #222, opened 2020-08-23
Added polygonize method

tyfoan:
### Summary

Added polygonize method of linestring and multilinestring


keithdoggett:
Thank you for submitting this and #221! I'm currently going through the issue backlog but will be able to review this and hopefully merge within the next few weeks.

test/geos_capi/multi_line_string_test.rb
```diff
@@ -12,6 +12,41 @@ class GeosMultiLineStringTest < Minitest::Test # :nodoc:
...
+      30.00098241341133 49.011536807005264, 30.0009793658682 49.01153814158362, 30.000978645524963 49.011540502810455,
+      30.00098067434918 49.01154250751114, 30.001038084302778 49.01156783398706, 30.001038084303158 49.011567833987236),
+      (30.001038084303158 49.011567833987236, 30.0011508215092 49.01161919123365, 30.00115085765015 49.01161920753694,
+      30.001182110397195 49.01163316771786, 30.001185695713303 49.01163365520981, 30.001188754941474 49.01163233224407,
+      30.001189496027187 49.01162997379613, 30.00118748485277 49.011627961412835, 30.001156250234796 49.011614009330685,
+      30.001043501660316 49.011562646908196, 30.001043463959558 49.011562641544636), (30.001043463959558 49.011562641544636,
+      30.001043463959554 49.011562641544636), (30.001062329555715 49.01154443246662, 30.001043463959558 49.011562641544636),
+      (30.001043463959558 49.011562641544636, 30.001038084303158 49.011567833987236),
+      (30.001038084303158 49.011567833987236, 30.00101922700232 49.01158603503694)
+    )")
```
keithdoggett, line 46: While this properly demonstrates the behavior, the geometry is a little complex and the coordinates are cumbersome due to their precision. I came up with this example that I think would be easier to debug/expand upon but still tests the method properly.

```rb
polygon1_wkt = "POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))"
polygon2_wkt = "POLYGON ((1 0, 1 1, 2 1, 2 0, 1 0))"

expected = @factory.parse_wkt("GEOMETRYCOLLECTION(#{polygon1_wkt}, #{polygon2_wkt})")

multi_line_string_wkt = "MULTILINESTRING( (1 1, 0 1, 0 0, 1 0), (1 -1, 1 0, 1 1, 1 2), (1 1, 2 1, 2 0, 1 0) )"
multi_line_string = @factory.parse(multi_line_string_wkt)
```

keithdoggett:
In addition to the above change, adding a test that confirms proper behavior on a failure should be included. Per the geos comments any of these should suffice:

```c
/*
 * The Polygonizer reports the follow kinds of errors:
 *
 * - Dangles - edges which have one or both ends which are
 *   not incident on another edge endpoint
 * - Cut Edges - edges which are connected at both ends but
 *   which do not form part of a polygon
 * - Invalid Ring Lines - edges which form rings which are invalid
 *   (e.g. the component lines contain a self-intersection)
*/
```

Also since this includes the same changes as #221, the comment from that could just be rolled into the changes here.

ext/geos_c_impl/geometry.c
```diff
@@ -764,6 +764,31 @@ static VALUE method_geometry_buffer_with_style(VALUE self, VALUE distance, VALUE
...
+                                                       rb_num2dbl(distance),
+                                                       RGEO_FACTORY_DATA_PTR(factory)->buffer_resolution,
+                                                       rb_num2int(joinStyle),
+                                                       rb_num2dbl(mitreLimit)),
+                                     Qnil);
+  }
+
+  return result;
+}
+
```
keithdoggett, line 791: `GeosOffsetCurve_r` only works on `linestrings` per the comments in `geos_c.h`

ext/geos_c_impl/geometry_collection.c
```diff
@@ -412,6 +412,35 @@ static VALUE method_multi_line_string_coordinates(VALUE self)
   return result;
 }
 
+
+static VALUE method_line_string_polygonize(VALUE self)
```
keithdoggett, line 416: Please change this to `method_multi_line_string_polygonize`

test/geos_capi/line_string_test.rb
```diff
@@ -22,4 +22,57 @@ def test_project_interpolate_round_trip
...
+      [30.000986002945037, 49.01153728085321],
+      [30.00098241341133, 49.011536807005264],
+      [30.0009793658682, 49.01153814158362],
+      [30.000978645524963, 49.011540502810455],
+      [30.00098067434918, 49.01154250751114],
+      [30.001038084302778, 49.01156783398706],
+      [30.001038084303158, 49.011567833987236],
+      [30.001043463959558, 49.011562641544636],
+      [30.001043463959554, 49.011562641544636]
+    ]
```
keithdoggett, line 39: Can you use simpler coordinates for this test?

test/geos_capi/line_string_test.rb
```diff
@@ -22,4 +22,57 @@ def test_project_interpolate_round_trip
...
+    
+  def test_offset_curve
+    factory = RGeo::Geos.factory(buffer_resolution: 4)
+    line_coordinates = [
+      [164.0, 16.0], [144.0, 16.0], [124.0, 16.0], [104.0, 16.0], [84.0, 16.0],
+      [64.0, 16.0], [44.0, 16.0], [24.0, 16.0], [20.0, 16.0], [18.0, 16.0],
+      [17.0, 17.0], [16.0, 18.0], [16.0, 20.0], [16.0, 40.0],
+      [16.0, 60.0], [16.0, 80.0], [16.0, 100.0], [16.0, 120.0],
+      [16.0, 140.0], [16.0, 160.0], [16.0, 180.0], [16.0, 195.0],
+    ]
```
keithdoggett, line 59: Can you use a simpler set of coordinates?

keithdoggett (changes requested):

keithdoggett:
I'd also like it if you try to do what I mentioned in this [comment](https://github.com/rgeo/rgeo/pull/222#issuecomment-713698022) and add some edge cases to the tests so we understand the impact on the library.

Additionally, please confirm that this PR includes all of the same changes as #211. If that is the case, then we should close that PR and just merge this.

test/geos_capi/line_string_test.rb
```diff
@@ -22,4 +22,41 @@ def test_project_interpolate_round_trip
     interpolated_point = line_string.interpolate_point location
     assert_equal point, interpolated_point
   end
+
+  def do_test_polygonize(expected_wkt, input_wkt)
```
keithdoggett, line 26: minor tweak, but can we change this to `assert_polygonize_equal`

test/geos_capi/multi_line_string_test.rb
```diff
@@ -12,6 +12,101 @@ class GeosMultiLineStringTest < Minitest::Test # :nodoc:
   include RGeo::Tests::Common::MultiLineStringTests
 
   def create_factory
-    RGeo::Geos.factory
+    @factory = RGeo::Geos.factory
+  end
+
+  def do_test_polygonize(expected_wkt, input_wkt)
```
keithdoggett, line 18: minor tweak, but can we change this to `assert_polygonize_equal`

keithdoggett (commented):
Thanks @tyfoan these tests look a lot better! Just change that method name and then it's good for me.

BuonOmo (commented):
Hey, thanks for the contribution!

@keithdoggett I definitely agree that some guidance may help (maybe in a CONTRIBUTING.md to follow github’s guidelines)

On the PRs content: I think we are missing some things.

On error, the polygonize geos method would return null, don’t we want to raise instead ? ([related doc](https://libgeos.org/doxygen/geos__c_8h.html#a9d98e448d3b846d591c726d1c0000d25)).

Also the method is undocumented, I think that is an issue and we should care to document it, either by linking to GEOS documentation directly if we do not change the interface, or adding our own doc.

Documenting C methods can be done thanks to the rdoc C parser https://ruby.github.io/rdoc/RDoc/Parser/C.html. I think we need to add documentation upfront even though it only works in v3-dev branch (yet) to make sure we don’t have too much documentation work afterwards.

keithdoggett:
@tyfoan could you just add a check if the result of `rgeo_wrap_geos_geometry` is `Qnil` and if it is raising an error with `rb_raise`? I think that would solve the issue for now and maybe eventually we can move that logic into the wrapping function. @BuonOmo

keithdoggett:
@tyfoan I agree since it may affect other methods and tests. So I guess the only thing left to do is document the function with a comment like @BuonOmo pointed out.

BuonOmo:
Hey @tyfoan,

Sorry for both the delay and lack of information in my previous message. I'll try to be more thorough here, and note that I can help further if needed.

## On returning `nil` vs raising

Indeed, it would be a tall task to ask from you to change `rgeo_wrap_geos_geometry`, and even though this is something we could think about, here it is really not the case. 

Since GEOS already gives us the information that the geometry is invalid by returning `null` before we have to cast to a rgeo value, that is already a good place to raise an error. Hence you'd be really focusing on a change that is really local to that PR.

Moreover, per GEOS documentation, the [`GEOSPolygonize_r`](https://libgeos.org/doxygen/geos__c_8h.html#a9d98e448d3b846d591c726d1c0000d25) method does in fact way more than what we check here. Among everything it does, there is:

> The Polygonizer reports the following kinds of errors:
>
> - Dangles - edges which have one or both ends which are not incident on another edge endpoint
> - Cut Edges - edges which are connected at both ends but which do not form part of a polygon
> - Invalid Ring Lines - edges which form rings which are invalid (e.g. the component lines contain a self-intersection)
>
> Errors are reported to output parameters "cuts", "dangles" and "invalid" (if not-null). Formed polygons are returned as a collection. NULL is returned on exception. All returned geometries must be destroyed by caller.

However, I've tried adding notice handlers to the PR and check for those, and I've seen nothing. Hence I think the tests only cover the successful part of `#polygonize`, not the failing on unfortunetaly... Here's a patch to add those handlers locally:

```diff
diff --git a/ext/geos_c_impl/factory.c b/ext/geos_c_impl/factory.c
index 8dc2025..cd73786 100644
--- a/ext/geos_c_impl/factory.c
+++ b/ext/geos_c_impl/factory.c
@@ -29,10 +29,25 @@ RGEO_BEGIN_C
 // NOP message handler. GEOS requires that a message handler be set
 // for every context handle.
 
-static void message_handler(const char* fmt, ...)
+static void notice_handler(const char* fmt, ...)
 {
+  va_list args;
+  va_start(args, fmt);
+  fprintf(stderr, "GEOS Notice -- ");
+  vfprintf(stderr, fmt, args);
+  fprintf(stderr, "\n");
+  va_end(args);
 }
 
+static void error_handler(const char* fmt, ...)
+{
+  va_list args;
+  va_start(args, fmt);
+  fprintf(stderr, "GEOS Error -- ");
+  vfprintf(stderr, fmt, args);
+  fprintf(stderr, "\n");
+  va_end(args);
+}
 
 // Destroy function for factory data. We destroy any serialization
 // objects that have been created for the factory, and then destroy
@@ -459,7 +474,7 @@ static VALUE cmethod_factory_create(VALUE klass, VALUE flags, VALUE srid, VALUE
   result = Qnil;
   data = ALLOC(RGeo_FactoryData);
   if (data) {
-    context = initGEOS_r(message_handler, message_handler);
+    context = initGEOS_r(notice_handler, error_handler);
     if (context) {
       data->geos_context = context;
       data->flags = NUM2INT(flags);

```

And I also think we should raise when an error message happens. Luckily, we're already working on that in the v3-dev branch (https://github.com/rgeo/rgeo/commit/a2f678da953024fd8cf1079954de7d6504dadbbe), so it is _kind of ok_ to keep it that way if we are sure we at least have tests covering those error messages.

SIDE: @keithdoggett we're more and more having conflicts between our v3-dev development and community's input, since we'd love more of that input I think we should release soon a `3.0.0-pre` version, WDYT?

## On documentation

I really have to insist on that since one day, someone will have to do it anyway hence we should start documenting new methods as they arrive. However, there are no examples available yet so I'll give it a try:

```diff
diff --git a/.yardopts b/.yardopts
index 133ac4f..bdf164f 100644
--- a/.yardopts
+++ b/.yardopts
@@ -1,5 +1,6 @@
 --markup rdoc
 --output-dir ./yardoc
 lib/**/*.rb
+ext/**/*.{c,h}
 -
 doc/*.md
diff --git a/Rakefile b/Rakefile
index ec5005f..091ced6 100644
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,8 @@
 require "rake/testtask"
 require "rake/extensiontask"
 require "bundler/gem_tasks"
+require "yard"
+
 
 # Build tasks
 
@@ -30,5 +32,12 @@ Rake::TestTask.new(:test) do |t|
   t.test_files = FileList["test/**/*_test.rb"]
 end
 
+YARD::Rake::YardocTask.new do |t|
+  # Runs a server to appreciate doc after having it running
+  t.after = proc do
+    exec "ruby", "-run", "-e", "httpd", File.join(__dir__, "yardoc")
+  end
+end
+
 task test: :compile
 task default: [:clean, :test]
diff --git a/ext/geos_c_impl/geometry_collection.c b/ext/geos_c_impl/geometry_collection.c
index be20de4..1358e4f 100644
--- a/ext/geos_c_impl/geometry_collection.c
+++ b/ext/geos_c_impl/geometry_collection.c
@@ -408,35 +408,6 @@ static VALUE method_multi_line_string_coordinates(VALUE self)
   return result;
 }
 
-
-static VALUE method_multi_line_string_polygonize(VALUE self)
-{
-  VALUE result;
-  RGeo_GeometryData* self_data;
-  const GEOSGeometry* self_geom;
-  GEOSGeometry* geos_polygon_collection;
-  VALUE factory;
-
-  result = Qnil;
-
-  self_data = RGEO_GEOMETRY_DATA_PTR(self);
-  self_geom = self_data->geom;
-
-  if (self_geom) {
-    factory = self_data->factory;
-    geos_polygon_collection = GEOSPolygonize_r(self_data->geos_context, &self_geom, 1);
-
-    result = rgeo_wrap_geos_geometry(
-        factory,
-        geos_polygon_collection,
-        Qnil
-      );
-  }
-
-  return result;
-}
-
-
 static VALUE method_multi_line_string_length(VALUE self)
 {
   VALUE result;
@@ -650,7 +621,7 @@ void rgeo_init_geos_geometry_collection()
   rb_define_method(geos_multi_line_string_methods, "closed?", method_multi_line_string_is_closed, 0);
   rb_define_method(geos_multi_line_string_methods, "hash", method_multi_line_string_hash, 0);
   rb_define_method(geos_multi_line_string_methods, "coordinates", method_multi_line_string_coordinates, 0);
-  rb_define_method(geos_multi_line_string_methods, "polygonize", method_multi_line_string_polygonize, 0);
+  rb_define_method(geos_multi_line_string_methods, "polygonize", method_line_string_polygonize, 0);
 
   // Methods for MultiPolygonImpl
   geos_multi_polygon_methods = rb_define_module_under(rgeo_geos_module, "CAPIMultiPolygonMethods");
diff --git a/ext/geos_c_impl/line_string.c b/ext/geos_c_impl/line_string.c
index ab6294f..61f5464 100644
--- a/ext/geos_c_impl/line_string.c
+++ b/ext/geos_c_impl/line_string.c
@@ -125,8 +125,20 @@ static VALUE method_line_string_coordinates(VALUE self)
   return result;
 }
 
-
-static VALUE method_line_string_polygonize(VALUE self)
+/**
+ * call-seq:
+ *   some.polygonize -> RGeo::Feature::GeometryCollection
+ *
+ * Polygonizes a set of line strings which contain linework that represents
+ * the edges of a planar graph.
+ *
+ * The edges must be correctly noded; that is, they must only meet at
+ * their endpoints. Polygonization will accept incorrectly noded input
+ * but will not form polygons from non-noded edges, and TODO: change this according to what we are going to do with the method.
+ *
+ * @see https://libgeos.org/doxygen/geos__c_8h.html#a9d98e448d3b846d591c726d1c0000d25 GEOSPolygonize
+ */
+VALUE method_line_string_polygonize(VALUE self)
 {
   VALUE result;
   RGeo_GeometryData* self_data;
diff --git a/ext/geos_c_impl/line_string.h b/ext/geos_c_impl/line_string.h
index 7cb0799..721f19a 100644
--- a/ext/geos_c_impl/line_string.h
+++ b/ext/geos_c_impl/line_string.h
@@ -18,6 +18,9 @@ RGEO_BEGIN_C
 */
 void rgeo_init_geos_line_string();
 
+// This method is also used by multi_line_strings.
+VALUE method_line_string_polygonize(VALUE);
+
 /*
   Determines whether the given GEOS line string is closed.
   Returns Qtrue if true, Qfalse if false, or Qnil on an error.
diff --git a/rgeo.gemspec b/rgeo.gemspec
index be39127..c615c8c 100644
--- a/rgeo.gemspec
+++ b/rgeo.gemspec
@@ -28,4 +28,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rake", "~> 13.0"
   spec.add_development_dependency "rake-compiler", "~> 1.0"
   spec.add_development_dependency "rubocop", "~> 1.8.1"
+  spec.add_development_dependency "yard", "~> 0.9"
 end

```

Also I'm not sure why we have two methods defined here, hence the diff removing one. The best imho would be to have a `Polygonizeable` module which would be included in both classes. However, this may be too much to ask and a compromise that would really be enough here is to just define the method in `line_string.{h,c}` which is for now included in the geometry_collection.c file anyway! And this is easier to document for sure :)

Finally, that diff contains what it takes to locally run `rake yard` and enjoy the documentation so you can try things and iterate easily!


Anyway, I hope the message was clearer. I'm sorry to bother you with all that but RGeo's codebase is not as sane as it could be, and we would not be able to keep it running if we were to accept PRs too quickly as we're only two maintainers and already [have a large backlog](#247)!

BuonOmo:
Closed by #313

PR #224, opened 2020-10-06
Add a simple fallback for Polygon#contains?

Quiwin:
Implement default Polygon#contains? and LineString#contains? which handle a Point as input.

### Summary

Add a fallback on `Polygon and LineString contains` so that if `geos` is not available, rgeo can still be used to determine if a point is inside a polygon.
Its using a raycasting algorithm (https://wrf.ecse.rpi.edu/Research/Short_Notes/pnpoly.html) with a border check.

### Other Information

Hello rgeo!
At Klaxit, we are planning to migrate from our internal geometry library to rgeo.
Since we will use rgeo in some of our open source projects we dont want to add mandatory c dependencies.

Happy to iterate if some changes are needed.

thank you!

### DIFF
--- a/lib/rgeo/impl_helper/basic_polygon_methods.rb
+++ b/lib/rgeo/impl_helper/basic_polygon_methods.rb
@@ -79,7 +79,44 @@ module RGeo
         ([@exterior_ring] + @interior_rings).map(&:coordinates)
       end
 
+      def contains?(rhs)
+        if Feature::Point === rhs
+          contains_point?(rhs)
+        else
+          raise(Error::UnsupportedOperation,
+            "Method Polygon#contains? is only defined for Point"
+          )
+        end
+      end
+
       private
+
+      def contains_point?(point)
+        ring_encloses_point?(@exterior_ring, point) &&
+          !@interior_rings.any? do |exclusion|
+            ring_encloses_point?(exclusion, point, on_border_return: true)
+          end
+      end
+
+      def ring_encloses_point?(ring, point, on_border_return: false)
+        # This is an implementation of the ray casting algorithm, greatly inspired
+        # by https://wrf.ecse.rpi.edu/Research/Short_Notes/pnpoly.html
+        # Since this algorithm does not handle point on edge, we check first if
+        # the ring is on the border.
+        # on_border_return is used for exclusion ring
+        return on_border_return if ring.contains?(point)
+        encloses_point = false
+        ring.points.each_cons(2) do |start_point, end_point|
+          next unless (point.y < end_point.y) != (point.y < start_point.y)
+
+          if point.x < (end_point.x - start_point.x) * (point.y - start_point.y) /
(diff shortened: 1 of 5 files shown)


test/simple_mercator/line_string_test.rb
```diff
@@ -17,4 +17,5 @@ def setup
 
   # These tests suffer from floating point issues
   undef_method :test_point_on_surface
+  undef_method :test_contains_point
```
Quiwin, line 20: There seems to be floating point issues, or should I add an epsilon for float comparison?
```
#<RGeo::Geographic::ProjectedLineStringImpl:0x3fdb07f02554 "LINESTRING (0.0 0.0, 1.0 2.0, 3.0 9.0, 0.0 0.0)">
#<RGeo::Geos::CAPILineStringImpl:0x3fdb07f020a4 "LINESTRING (0.0 -7.081154551613622E-10, 111319.49079327357 222684.20850554318, 333958.4723798207 1006021.0627551326, 0.0 -7.081154551613622E-10)">
#<RGeo::Geos::CAPIPointImpl:0x3fdb07f0784c "POINT (55659.74539663678 111325.14286638486)">
```

lib/rgeo/impl_helper/basic_polygon_methods.rb
```diff
@@ -79,8 +79,45 @@ def coordinates
...
+          end
+      end
+
+      def ring_encloses_point?(ring, point, on_border_return: false)
+        # This is an implementation of the ray casting algorithm, greatly inspired
+        # by https://wrf.ecse.rpi.edu/Research/Short_Notes/pnpoly.html
+        # Since this algorithm does not handle point on edge, we check first if
+        # the ring is on the border.
+        # on_border_return is used for exclusion ring
+        return on_border_return if ring.contains?(point)
```
keithdoggett, line 107: I think that we should get rid of this optional hash parameter `on_border_return` and just make it false in the code. We can always change it if we determine that should be true later on or some optional parameter, but as it stands this isn't configurable by the user so I think it's cleaner just to make the statement:

```rb
return false if ring.contains?(point)
```

and get rid of the parameter.

keithdoggett:
Thanks for submitting this. I think it looks great except for the one tweak that I noted above.

lib/rgeo/impl_helper/basic_polygon_methods.rb
Quiwin: This parameter is used to change the border behavior for exclusion rings. I'm not sure how we could remove it.

keithdoggett: Apologies, you're correct. I somehow missed that it was used with both `true` and `false`. 

This looks good then.

keithdoggett (approved):

BuonOmo (approved):
@keithdoggett I'm merging it then !

PR #228, opened 2020-10-30
Standardize Validation

### LINKED ISSUE #218: Different Validation Behavior (is_simple) in Geos 3.8.x
RGeo has different exception throwing behavior and returns different `is_simple` values when backed by different versions of the underlying `geos` library.

Specifically the `RGeo::Geographic.simple_mercator_factory` will throw an exception when parsing a self intersecting "hour glass" polygon in `3.8` while it will parse that polygon and return `#is_simple?` `true` in `3.7`.

I think this is the root cause of this stackoverflow issue: https://stackoverflow.com/questions/58996305/getting-invalidgeometry-linearring-failed-ring-test-after-upgrading-rgeo-gem

### Steps to reproduce
With `geos-config --version` `3.8.1`:
```
RGeo::Geographic.simple_mercator_factory.parse_wkt('POLYGON((-93.572755591195 44.615834475162, -93.571098455347 44.613948476312, -93.571098455347 44.613948476313, -93.571098455347 44.613948476314, -93.571325365812 44.615739119667, -93.572627021193 44.613957229694, -93.572755591195 44.615834475162))')
```
This is a self-intersecting "hour glass" polygon:
geojson_io

### Expected behavior
With `geos-config --version` `3.5.2`, `3.6.4`, or `3.7.4` (built from tarball with cmake):
This Polygon will be parsed. It returns `true` for `#is_simple?`.

### Actual behavior
`RGeo::Error::InvalidGeometry (LinearRing failed ring test)`

### System configuration

**Ruby version**:
`ruby 2.5.7p206 (2019-10-01 revision 67816) [x86_64-darwin18]`

**OS**:
`Darwin jdalt-mbp.local 18.7.0 Darwin Kernel Version 18.7.0: Thu Jan 23 06:52:12 PST 2020; root:xnu-4903.278.25~1/RELEASE_X86_64 x86_64`

Also replicated on travis-ci with Ubuntu xenial with a `geos` 3.8.1 built from source/tarball.

`rgeo` is version `2.1.1`

### Speculation
I think this is a difference in how `is_simple` is calculated in the CAPI. I believe all factories that use the `CAPI` to calculate `is_simple` are affected. I don't think the behavior of `RGeo::Geographic.spherical_factory` is affected.

### Additional Thoughts
This is a breaking change, but for my own purposes I prefer the new behavior. I like it because it's more consistent with the implementation of `RGeo::Geographic.spherical_factory` which I converted my app to in order to avoid the issues outlined in https://github.com/rgeo/rgeo/issues/212. I tried #213 but it caused too much of a performance degradation to be viable for me.

However, I can't easily upgrade the underlying library in my production environment, so I'd like to get more insight into what's happening and what other mitigation options might be available to me. Ideally I'd be able to induce the 3.8 behavior in earlier versions of the library, or I'd be alerted to some other factory that's performant and consistent across versions.

I think others need to be aware that there are serious, breaking changes that exist in `geos` `3.8.x`. This will become the default version of the library in Ubuntu `focal`. I imagine others who frequently hop back and forth between `macOS` (with the library installed and updated by Homebrew) and Linux ci and/or production environments will be similarly confused by this behavior.

keithdoggett:
### Summary

PR to fix #218 where `is_simple?` returns inconsistent results across libgeos versions.

### Other Information

I personally tried using geos versions 3.5.2, 3.6.4, 3.7.3, and 3.8.1 across multiple Rubies and did not have a failing test case. I'm using Ubuntu Bionic so that may have something to do with it, but this is inconsistent with what is described in #218 .

I also looked at the libgeos docs and there is a note in there that suggests `GeosIsSimple` will always return `true` for polygons and that `GeosIsValid` would be the appropriate function to check for self-intersections. This is also described here https://postgis.net/docs/using_postgis_dbmanagement.html#OGC_Validity where it says "By definition, a POLYGON is always simple."

### DIFF
--- a/test/geos_capi/polygon_test.rb
+++ b/test/geos_capi/polygon_test.rb
@@ -11,10 +11,6 @@ require "test_helper"
 class GeosPolygonTest < Minitest::Test # :nodoc:
   include RGeo::Tests::Common::PolygonTests
 
-  def assert_close_enough(p1, p2)
-    assert((p1.x - p2.x).abs < 0.00000001 && (p1.y - p2.y).abs < 0.00000001)
-  end
-
   def setup
     @factory = RGeo::Geos.factory
   end
@@ -117,9 +113,12 @@ class GeosPolygonTest < Minitest::Test # :nodoc:
     # 4.3 -> 4.29999999999999, for example, and throws an error
     # iterating through points and using assert_in_delta instead
     # of assert_equal
-    buffered_points = buffered_line_string.exterior_ring.points
-    polygon.exterior_ring.points.each_with_index do |pt, idx|
-      assert_close_enough(pt, buffered_points[idx])
+    b_coords = buffered_line_string.exterior_ring.coordinates
+    p_coords = polygon.exterior_ring.coordinates
+    p_coords.zip(b_coords).each do |p_coord, b_coord|
+      p_coord.zip(b_coord).each do |pt1, pt2|
+        assert_in_delta(pt1, pt2, 1e-7)
+      end
     end
   end
 
@@ -154,15 +153,5 @@ class GeosPolygonTest < Minitest::Test # :nodoc:
     outer_ring = @factory.linear_ring(points_arr)
     polygon = @factory.polygon(outer_ring)
     polygon.invalid_reason
-  end
-
-  def test_self_intersecting_polygon
-    # issue 218
-    polygon_coordinates = [[0, 0], [1, 1], [0, 1], [1, 0], [0, 0]]
-    points_arr = polygon_coordinates.map{ |v| @factory.point(v[0], v[1]) }
-    outer_ring = @factory.linear_ring(points_arr)
-    polygon = @factory.polygon(outer_ring)
-
-    refute(polygon.is_simple?)
   end
 end if RGeo::Geos.capi_supported?
--- a/test/geos_capi/polygon_test.rb
+++ b/test/geos_capi/polygon_test.rb
@@ -11,6 +11,10 @@ require "test_helper"
 class GeosPolygonTest < Minitest::Test # :nodoc:
   include RGeo::Tests::Common::PolygonTests
 
+  def assert_close_enough(p1, p2)
+    assert((p1.x - p2.x).abs < 0.00000001 && (p1.y - p2.y).abs < 0.00000001)
+  end
+
   def setup
     @factory = RGeo::Geos.factory
   end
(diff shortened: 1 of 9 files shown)


keithdoggett:
After trying again, I was able to get tests to fail locally by switching libgeos versions. The issue is that in older versions of libgeos (<3.8), only `LineStrings`, `MultiLineStrings` and `MulitPoints` had a simplicity check while everything else returned `true` by default. This was modified by this [commit](https://github.com/libgeos/geos/commit/dddda853c784296ba9f3742e1659e8140c0e8a33#diff-80b3a73336b1e5fbb666a8ef8df87b775c8cb6fd51443391be7e63547a0af998L111)  and released in 3.8.0.

The simplicity check that the `Geographic` module implements assumes that `LinearRing#is_simple?` will return `false` given a self-intersection, but this is inconsistent with older versions of libgeos. To fix this, `ProjectedLinearRingImpl#is_simple?` will now rely on `CAPILinearRingImpl#valid?` instead of `CAPILinearRingImpl#is_simple?` since this is consistent with the `Geographic` definition. 

This will be a breaking change since users with older versions of libgeos will now have more strict validation requirements on projected factories.

keithdoggett:
The reason for overriding `is_simple?` in `ProjectedLinearRingMethods` instead of `validate_geometry` is because the idea behind projection factories is that they can be almost swappable with the spherical factory. This means that the only real difference should be that the spherical factory uses a spherical math backend while the proejction factories use a PCS as their backend, but the methods should have the same behavior. In this case, `is_simple?` in the spherical factory was equivalent to `valid?` in the libgeos API so this will make the behavior between each factory more consistent.

The only issue with this could be if there's a case where a `LinearRing` is simple but not valid or vice versa. I'll look into this but I think it should be ok.

test/geos_capi/polygon_test.rb
```diff
@@ -109,7 +109,17 @@ def test_buffer_with_style
...
+    # 4.3 -> 4.29999999999999, for example, and throws an error
+    # iterating through points and using assert_in_delta instead
+    # of assert_equal
+    b_coords = buffered_line_string.exterior_ring.coordinates
+    p_coords = polygon.exterior_ring.coordinates
+    p_coords.zip(b_coords).each do |p_coord, b_coord|
+      p_coord.zip(b_coord).each do |pt1, pt2|
+        assert_in_delta(pt1, pt2, 1e-7)
+      end
+    end
```
BuonOmo, line 122: This one is not satisfying, but this PR is not the place to fix floating point errors!

test/simple_mercator/polygon_test.rb
```diff
@@ -15,6 +15,14 @@ def setup
     @factory = RGeo::Geographic.simple_mercator_factory
   end
 
+  def test_is_simple_validation_behavior
+    # issue 218
```
BuonOmo, line 19: I always find it easier to show an url in that case:
```suggestion
    # See https://github.com/rgeo/rgeo/issues/218
```

BuonOmo (approved):
LGTM, however it is definitely a breaking change. I guess you could add this change to the History.md file and we'll start a dev-3 branch (or the naming your prefer!)

PR #230, opened 2020-11-16
Add fallback for MultiPolygon#contains? on point

Quiwin:
### Summary

Add a fallback on `MultiPolygon#contains?`.

Since we have a fallback on `Polygon#contains?`, we might as well use it for a simple `MultiPolygon#contains?` on a point.

### DIFF
--- a/lib/rgeo/impl_helper/basic_geometry_collection_methods.rb
+++ b/lib/rgeo/impl_helper/basic_geometry_collection_methods.rb
@@ -177,6 +177,12 @@ module RGeo
       def coordinates
         @elements.map(&:coordinates)
       end
+
+      def contains?(rhs)
+        return super unless Feature::Point === rhs
+
+        @elements.any? { |poly| poly.contains?(rhs) }
+      end
     end
   end
 end
--- a/test/common/multi_polygon_tests.rb
+++ b/test/common/multi_polygon_tests.rb
@@ -205,6 +205,61 @@ module RGeo
             end
           end
         end
+
+        def test_contains_point
+          geom1 = @factory.multi_polygon([@poly1, @poly2])
+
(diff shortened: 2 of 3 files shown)


lib/rgeo/impl_helper/basic_geometry_collection_methods.rb
```diff
@@ -177,6 +177,19 @@ def boundary
       def coordinates
         @elements.map(&:coordinates)
       end
+
+      def contains?(rhs)
+        unless Feature::Point === rhs
+          raise(Error::UnsupportedOperation,
+            "Method MultiPolygon#contains? is only defined for Point"
+          )
```
BuonOmo, line 185:
```suggestion
          super
```

There may be a super method handling that

lib/rgeo/impl_helper/basic_geometry_collection_methods.rb
```diff
@@ -177,6 +177,19 @@ def boundary
...
+        end
+
+        @elements.each do |poly|
+          return true if poly.contains?(rhs)
+        end
+        false
```
BuonOmo, line 191:
```suggestion
        @elements.any? { |polygon| polygon.contains?(rhs) }
```

Some rubyification/simplification :)

BuonOmo (approved):
LGTM! @keithdoggett I'll let your review and merge if you are also ok with the feature :)

@quiwin I've proposed 2 changes, could you review those ?

keithdoggett:
Looks good! Thanks for submitting this. What are your thoughts on extending this to include `MultiLineStrings` as well? In your last PR (#224), you added `LineString#contains?` so it should be simple enough to make it work for `MultiLineStrings`.

keithdoggett (approved):

PR #232, opened 2020-11-18
Add `MultiLineString#contains?` point fallback

keithdoggett:
### Summary

Adds the `contains?` method for `MutliLineString` cartesian and spherical implementations.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,9 +1,11 @@
 ### Current
+
 * Add SphericalPolygonMethods#centroid #208 (allknowingfrog)
 * Expand gemspec
 * Drop Ruby 2.3 support
 * Add a simple fallback for `Polygon#contains?` (Quiwin) #224
 * Add `ccw?` method for linear rings, with geos support #229
+* Add a simple fallback for `MultiPolygon/LineString#contains?` (Quiwin) #230/231
 
 ### 2.1.1 / 2019-8-26
 
--- a/History.md
+++ b/History.md
@@ -5,7 +5,7 @@
 * Drop Ruby 2.3 support
 * Add a simple fallback for `Polygon#contains?` (Quiwin) #224
 * Add `ccw?` method for linear rings, with geos support #229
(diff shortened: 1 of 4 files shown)


History.md
```diff
@@ -1,9 +1,11 @@
 ### Current
+
 * Add SphericalPolygonMethods#centroid #208 (allknowingfrog)
 * Expand gemspec
 * Drop Ruby 2.3 support
 * Add a simple fallback for `Polygon#contains?` (Quiwin) #224
 * Add `ccw?` method for linear rings, with geos support #229
+* Add a simple fallback for `MultiPolygon/LineString#contains?` (Quiwin) #230/231
```
BuonOmo, line 8:
```suggestion
* Add a simple fallback for `MultiPolygon/LineString#contains?` (Quiwin) #230 #232
```

BuonOmo (approved):
One little issue, LGTM otherwise !

keithdoggett:
I don't think we need to worry about this now, but technically the `LineString#contains?` implementation is inconsistent with the GEOS implementation. This snippet from the [postgis docs](url) describes it:

>Geometry A contains Geometry B if and only if no points of B lie in the exterior of A, and at least one point of the interior of B lies in the interior of A. An important subtlety of this definition is that A does not contain its boundary, but A does contain itself.

So a point that lies on either terminal point of the `LineString` should return `false` for `contains?` which it does in GEOS, but not for the cartesian implementation. I can open up an issue/PR for it soon, but I don't think that fix belongs in this PR.

BuonOmo:
Indeed, let's not delay this minor even more ! However this definition makes me think that a LineString should never contain anything ? Since it has only borders... I don't quite grasp why it is important to not match first and last point but to to match every other points...

keithdoggett:
It feels especially weird for linestrings because the boundary of a geometry is always one dimension less than the dimension of the geom, so the boundaries are the terminal points. This makes sense by the definition, but intuitively it seems strange. 

Regardless, we can worry about that later. I'll merge this in now and get a release out.

Issue #234, opened 2020-11-24
Move to Github actions


BuonOmo:
Travis recently changed plans pricing which makes it impossible to continue Open Source on travis. We should move to github actions

PR #240, opened 2021-01-14
Use Github Actions for CI

mlarraz:
### Summary

Addresses https://github.com/rgeo/rgeo/issues/234.

Drops support for Ruby 2.4, adds it for Ruby 3.0.

See action on fork: https://github.com/mlarraz/rgeo/runs/1714762543


keithdoggett (approved):
Looks good! Thanks for submitting this.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,30 @@
...
+      - master
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
```
BuonOmo, line 15: I guess we could include `head` as well, to be sure we don't miss on ruby updates !

.github/workflows/CI.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
BuonOmo, line 11: Shouldn't we also have builds working on mac, maybe windows ? (@keithdoggett I don't the app state on windows support)

BuonOmo (commented):
LGTM as well! Thanks for the contribution :)

I guess we could already merge it since it is better than actual configuration. However, I've adressed two possible improvement, and I guess now is a good time to discuss those!

.github/workflows/CI.yml
keithdoggett: yeah I think we should include those. I've used it on windows before.

mlarraz: I've updated, but specs are failing on Windows. I set them as allowed failures for now. Ideally someone else can tackle getting things working on Windows after this is merged

keithdoggett: Yeah we can deal with fixing that once it's merged. Are you able to add `macos-latest` as well? It may be easier to break the task up into different jobs for each OS since installing libgeos will be different for each.

Something like this:
```yml
name: CI

on:
  push:
    branches:
      - master
      - github_actions
  pull_request:

jobs:
  ubuntu:
    strategy:
      fail-fast: false
      matrix:
        ruby: [ruby-head, 3.0, 2.7, 2.6, 2.5, jruby]
    runs-on: ubuntu-latest
    continue-on-error: ${{ matrix.ruby == 'ruby-head' }}
    steps:
      - uses: actions/checkout@v2
      - uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
          bundler-cache: true
      - run: |
          sudo apt-get install libgeos-dev -y
      - run: |
          bundle exec rake
  macos:
    strategy:
      fail-fast: false
      matrix:
        ruby: [ruby-head, 3.0, 2.7, 2.6, 2.5, jruby]
    runs-on: macos-latest
    continue-on-error: true
    steps:
      - uses: actions/checkout@v2
      - uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
          bundler-cache: true
      - run: |
          brew install geos
      - run: |
          bundle exec rake
  windows:
    strategy:
      fail-fast: false
      matrix:
        ruby: [ruby-head, 3.0, 2.7, 2.6, 2.5, jruby]
    runs-on: windows-latest
    continue-on-error: true
    steps:
      - uses: actions/checkout@v2
      - uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
          bundler-cache: true
      # figure out how to install on windows. Probably need mingw
      - run: |
          sudo apt-get install libgeos-dev -y
      - run: |
          bundle exec rake
```
I haven't tested this out, but I think something like this would make it easier to figure out how to build for different operating systems.

mlarraz: Updated to add macOS too. Also set up scaffolding for OS-specific setup (Windows is blank right now).

BuonOmo (approved):
Nice work, thank you for your reactivity! I guess you're right: If one want CI on windows CI and is working on windows, he'll have to contribute. I'll open one issue once this is merged 🙂

keithdoggett:
Looks good! Thanks for the help.

PR #243, opened 2021-01-19
Add documentation to the codebase

BuonOmo:
### Summary

In the context of improving our documentation (#235), we want to have it living with our codebase. This PR is the first step towards that goal:

1. Add yard options that allows rendering, to test the result try `yard server`
2. Move files from wiki to docs (@keithdoggett that was done on January 18, check if your latest changes are applied!)
3. Attest these changes in a documentation section in the readme
4. Ensure compatibility table is always up to date with a test.

### Next step

@keithdoggett as discuss we will soon come with a CSS/HTML proposition for the github pages. The idea is to have only one page with an example, useful links and sponsors.

@joakimklaxit I couldn't assign you, so here it is, you are at least mentionned!


.yardopts
```diff
@@ -0,0 +1,5 @@
+--markup markdown
```
BuonOmo, line 1: @keithdoggett this changes things a bit in fact: it favors markdown in stead of rdoc. Hence the next code:

```
# this is `markdown` versus +rdoc+
```

Would render the `markdown` part but not the +rdoc+ part. I don't think we can use both. And I'm leaning towards markdown since it is more powerful and well known. What is your opinion on that ?

README.md
```diff
@@ -108,25 +108,48 @@ ActiveRecord connection adapter for SpatiaLite, based on sqlite3 (*not maintaine
...
 
-### Acknowledgments
+You can see more in-depth documentation in the `doc` folder. Factories and
+methods are documented inline, you should consider checking
+https://rubydoc.info/gems/rgeo with the version you are currently using. Or
+generate documentation locally if you're working on RGeo: `yardoc server`.
+
+Here's the current list of available topics:
+
+- [An introduction to Spatial Programming With RGeo](/docs/file/doc/An-Introduction-to-Spatial-Programming-With-RGeo.md)
```
BuonOmo, line 127: These links work on rubydoc but not on github. There may be solutions to that using handler plugins (@JoakimKlaxit did you see if it was possible to do so with markdown files?).

Another solution is to have a full link, forcing one to github or to rubydoc (the latter is cleaner since it contains references to the documentation in ruby code blocs)

Otherwise which one should we favor? Or do you have any clue on how to solve that ?

doc/Factory-Compatibility.md
```diff
@@ -0,0 +1,768 @@
...
+| projection | `RGeo::Geographic.simple_mercator_factory`                                                   | Compound factory that uses a spherical factory to create geometries, but a geos or cartesian factory for computations on a projected coordinate system |
+
+For more information about choosing a factory, see the [Choosing a Fatory doc](https://github.com/rgeo/rgeo/wiki/Which-factory-should-I-use%3F-(draft)).
+
+## Tested Methods
+
+The table tests all of the methods defined in the [OGC SFA v1.1](https://www.ogc.org/standards/sfa) for each geometry type. In the case of methods that require another geometry (ex. `LineString#interects?(Polygon)`), all geometry types will be tested as the parameter.
+
+_Note: This list is not exhaustive of all the methods defined by each factory. The geos factory in particular has a good amount of non-standard methods implemented. These can be found in the YARD docs._
+<!-- TODO: this is not true, C methods are not listed in documentation, we should do something about it. -->
```
BuonOmo, line 26: @keithdoggett @JoakimKlaxit FYI. We should dig into that

.yardopts
keithdoggett: So this means that it will not generate documentation based on the comments in the code?

README.md
keithdoggett: If we had to choose one, I'd prefer the former, but I've seen this handled before and I think it can work for both.

When I have more time, I can research it.

doc/Factory-Compatibility.md
keithdoggett: Yes, agreed. I've gotten this to work on my package before, so it may just be touching the format up. [code](https://github.com/keithdoggett/spatial_stats/blob/master/ext/spatial_stats/csr_matrix.c#L112) : [corresponding docs](https://keithdoggett.github.io/spatial_stats/SpatialStats/Weights/CSRMatrix.html#initialize-instance_method)

test/factory_compatibility_table_test.rb
```diff
@@ -0,0 +1,230 @@
...
+      :envelope,
+      :as_text,
+      :as_binary,
+      :is_empty?,
+      :is_simple?,
+      :boundary,
+      :convex_hull,
+      :buffer
+    ].freeze
+  end
```
keithdoggett, line 63: We may also want to consider a way to add methods for specific geometry classes. For example, `Polygon#area`, `LineString#length`, etc.

We don't need to change this now, but we should make sure that these methods could be extended to account for those.

keithdoggett (commented):
This is awesome! Great work. I left just a few notes.

.yardopts
BuonOmo: @keithdoggett yes it will ! But it will use the markdown syntax rather than the rdoc syntax. Yardoc specific syntax will still be available either way

doc/Factory-Compatibility.md
BuonOmo: @keithdoggett nice ! I'd love to know the solution for my own projects !

.yardopts
keithdoggett: Ok, so this means that we will have to change the RDoc specific syntax to markdown syntax.

For example in this snippet:
``` 
# [<tt>:has_z_coordinate</tt>]
#   Set to true if geometries created by this factory include a Z
#   coordinate, and the Point#z method is available.
# [<tt>:has_m_coordinate</tt>]
#   Set to true if geometries created by this factory include a M
#   coordinate, and the Point#m method is available.
# [<tt>:is_cartesian</tt>]
#   Set to true if this Factory guarantees that it operates in
#   Cartesian geometry. If false or nil, no such guarantee is made,
#   though it is possible the geometries may still be Cartesian.
# [<tt>:is_geographic</tt>]
#   Set to true if this Factory's coordinate system is meant to be
#   interpreted as x=longitude and y=latitude. If false or nil, no
#   information is present about whether the coordinate system is
#   meant to be so interpreted.
```
The `<tt>` tags would have to be changed to the equivalent in markdown?

BuonOmo: This is kind of wrong any way and would need to use rather yard specific syntax. However i don't know how does this render using the yardopts we added. Did you take a look at that part of the doc ?

keithdoggett: Looks like it works, but it probably needs an extra line between each option. It's rendering everything on the same line for me. Although that might be an issue with my setup since I was seeing the same thing with the table.

keithdoggett:
Aside from figuring out that formatting issue where there are missing newlines, we should look at the failing CI as well.

It looks like the Ubunutu builds are breaking for 2 reasons:

1) It has the C Extension but no FFI.
2) It has FFI but no C Extension.

Since the compatibility table requires both, we should create a way to skip that test for the CI and for FFI users.

I propose adding a new rake task and modifying the test task.

```rb
# Rakefile

# line 31
Rake::TestTask.new(:test) do |t|
  t.libs << "test"
  t.libs << "lib"
  t.test_files = FileList["test/**/*_test.rb"].exclude("test/factory_compatibility_table_test.rb")
end

Rake::TestTask.new(:test_factory_compatibility) do |t|
  t.libs << "test"
  t.libs << "lib"
  t.test_files = FileList["test/factory_compatibility_table_test.rb"]
end

task test: :compile
task test_factory_compatibility: :compile
task default: [:clean, :test, :test_factory_compatibility]
```

Then change the CI to run `bundle exec rake test`.

test/factory_compatibility_table_test.rb
```diff
@@ -0,0 +1,230 @@
...
+        *factories.map { |factory| handled_per_factory_per_description[factory][description] ? "✅" : "❌" }
+      ]
+    end
+
+    markdown = ""
+    markdown += "| #{titles.zip(sizes).map { |title, size| title.to_s.center(size) } * ' | '} |\n"
+    markdown += "| #{ '-' * (sizes.first - 1) + ':' } | #{sizes[1..-1].map { |size| ':' + '-' * (size - 2) + ':' } * ' | '} |\n"
+    rows.each do |row|
+      markdown += "| #{row.first.rjust(sizes.first)} | #{row[1..-1].zip(sizes[1..-1]).map { |cell, size| cell.center(size) } * ' | '} |\n"
+    end
```
keithdoggett, line 132: This looks correct and should be working, but on my computer, it is rendering on one line. I'm on the latest version of yard and checked on both chrome and firefox and saw the same issue.

keithdoggett: I installed the `redcarpet` gem and didn't change anything else (just ran `gem install redcarpet`), and it rendered properly. I thought I would have to set it with the -M flag in `.yardopts`, but I guess it automatically detects one.

Do you have one installed or did it work out of the box for you?

Running `yard markups | grep markdown` shows this list of available providers.

```
[markdown]
  Providers: redcarpet rdiscount kramdown bluecloth maruku rpeg-markdown rdoc
  Extensions: .markdown .md .mdown .mkd
```

Will we need to add one of these as a dev dependency?

BuonOmo: @keithdoggett good catch! I do have all of these as well:

```
[markdown]
  Providers: redcarpet rdiscount kramdown bluecloth maruku rpeg-markdown rdoc
  Extensions: .markdown .md .mdown .mkd
```

We could also switch back to the html table ? (I did the markdown one for readability as a file)

I don't see documentation about markup provider in yard documentation. If you could uninstall back redcarpet and see if adding it to dev dependencies works ? If you cannot reproduce I'll do that one next week

README.md
BuonOmo: I've done three things:

- links are hardcoded to github
- there is a new link to the index page on rubydoc.info to make sure a user always has this list up to date
- add a test that check if readme is up to date with doc folder

Here it is https://github.com/klaxit/rgeo/commit/4954cb04abcb1d98bb6069e9f6c877322b52329b

doc/Installing-GEOS.md
```diff
@@ -0,0 +1,96 @@
...
+
+You can then check in an irb session that it is correctly binded to your `rgeo` gem:
+
+```ruby
+require "rgeo"
+
+puts "Yay" if RGeo::Geos.capi_supported?
+```
+
+If you need to install it on heroku, [there is a dedicated wiki page](https://github.com/rgeo/rgeo/wiki/Enable-GEOS-and-Proj4-on-Heroku).
```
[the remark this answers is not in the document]
BuonOmo: fixed

test/factory_compatibility_table_test.rb
BuonOmo: @keithdoggett this will come later indeed !

test/factory_compatibility_table_test.rb
BuonOmo: I couldn't make it _not work_. I'd say we publish as-is and see if rubydoc has the correct parser

BuonOmo:
> Aside from figuring out that formatting issue where there are missing newlines, we should look at the failing CI as well.
> 
> It looks like the Ubunutu builds are breaking for 2 reasons:
> 
> 1. It has the C Extension but no FFI.
> 2. It has FFI but no C Extension.
> 
> Since the compatibility table requires both, we should create a way to skip that test for the CI and for FFI users.
> 
> I propose adding a new rake task and modifying the test task.
> 
> ```ruby
> # Rakefile
> 
> # line 31
> Rake::TestTask.new(:test) do |t|
>   t.libs << "test"
>   t.libs << "lib"
>   t.test_files = FileList["test/**/*_test.rb"].exclude("test/factory_compatibility_table_test.rb")
> end
> 
> Rake::TestTask.new(:test_factory_compatibility) do |t|
>   t.libs << "test"
>   t.libs << "lib"
>   t.test_files = FileList["test/factory_compatibility_table_test.rb"]
> end
> 
> task test: :compile
> task test_factory_compatibility: :compile
> task default: [:clean, :test, :test_factory_compatibility]
> ```
> 
> Then change the CI to run `bundle exec rake test`.

I ended up skipping in the test when `capi_supported?` or `ffi_supported?` returns false. This feels better because it keeps the complexity in that one specific method. What do you think ?

test/factory_compatibility_table_test.rb
keithdoggett: Agreed. I dug through the YARD code a little bit to see what might be causing it but I think we should just publish and if it's an issue we can deal with it then.

keithdoggett:
> > Aside from figuring out that formatting issue where there are missing newlines, we should look at the failing CI as well.
> > It looks like the Ubunutu builds are breaking for 2 reasons:
> > 
> > 1. It has the C Extension but no FFI.
> > 2. It has FFI but no C Extension.
> > 
> > Since the compatibility table requires both, we should create a way to skip that test for the CI and for FFI users.
> > I propose adding a new rake task and modifying the test task.
> > ```ruby
> > # Rakefile
> > 
> > # line 31
> > Rake::TestTask.new(:test) do |t|
> >   t.libs << "test"
> >   t.libs << "lib"
> >   t.test_files = FileList["test/**/*_test.rb"].exclude("test/factory_compatibility_table_test.rb")
> > end
> > 
> > Rake::TestTask.new(:test_factory_compatibility) do |t|
> >   t.libs << "test"
> >   t.libs << "lib"
> >   t.test_files = FileList["test/factory_compatibility_table_test.rb"]
> > end
> > 
> > task test: :compile
> > task test_factory_compatibility: :compile
> > task default: [:clean, :test, :test_factory_compatibility]
> > ```
> > 
> > 
> > Then change the CI to run `bundle exec rake test`.
> 
> I ended up skipping in the test when `capi_supported?` or `ffi_supported?` returns false. This feels better because it keeps the complexity in that one specific method. What do you think ?

That also works. We should try to figure out getting both the CAPI and FFI to work on the CI. It seems like only one or the other currently works. We might just have to install the FFI gem on the MRI builds as well as the Jruby builds.

.gitignore
```diff
@@ -13,3 +12,4 @@ tmp
 Makefile*
 mkmf.log
 Gemfile.lock
+yardoc
```
keithdoggett, line 15: should this be `.yardoc`?

keithdoggett (approved):

BuonOmo:
> That also works. We should try to figure out getting both the CAPI and FFI to work on the CI. It seems like only one or the other currently works. We might just have to install the FFI gem on the MRI builds as well as the Jruby builds.

Or even do like the rubocop pattern: run a specific build just for this test !

Also I haven't committed it but locally the support changed after rebase. Let me add the commit for that as well :)

keithdoggett:
Looks good! I left one comment about the `yardoc` path on the .gitignore.

Also, what are your thoughts on placing the documentation_links_test and factory_compatibility_table_test in a test/docs directory? I'm also going to add another one to test the examples in the Examples.md file I just finished. Might be good to group them in a specific docs test directory.

.gitignore
BuonOmo: Both in fact! (the yardoc folder is used for generated doc in our `.yardopts` file)

PR #244, opened 2021-01-19
Update Rubocop and add to CI

mlarraz:
### Summary
Noticed Rubocop is installed and configured but not running and out of date, so addressing in a minimally invasive way.


keithdoggett:
This looks good, thanks for submitting this! My only question is if this should go in the same workflow as the tests or be a separate workflow. I don't think it really matters, I'm just not sure what best practice is.

keithdoggett (approved):

mlarraz:
> This looks good, thanks for submitting this! My only question is if this should go in the same workflow as the tests or be a separate workflow. I don't think it really matters, I'm just not sure what best practice is.

I think in this case since the triggers are the same it makes sense to group everything under the same workflow.

BuonOmo (approved):
LGTM as well! As for doing it in the test workflow or only once, I'd say the former has no real purpose (code should be the same whatever the version of ruby/os, one target is enough) and the latter is faster.

About the `.rubocop_todo`, if you are open to fix some of those in another PR we'd be glad :)

PR #246, opened 2021-02-02
Add Examples to Documentation

keithdoggett:
### Summary

* Add `doc/Examples.md` document. Demonstrates creating factories and features, then shows examples of most analysis and predicate methods.
* Add `ExamplesTest` that writes example code into a `Tempfile`, then loads it to confirm that the code works.


BuonOmo (commented):
Here's a bit different way to test examples one by one. I think this is better since it ensures tests are working on their own. And it gives a better feedback. 

One issue left though: errors are given one by one and not aggregated, which is a bit frustrating IMHO..

image

```patch
diff --git a/test/docs/examples_test.rb b/test/docs/examples_test.rb
index a6085c1..33fc554 100644
--- a/test/docs/examples_test.rb
+++ b/test/docs/examples_test.rb
@@ -13,17 +13,17 @@ class ExamplesTest < MiniTest::Test
       skip 'Examples can only be run with FFI and CAPI support'
     end
 
-    file = Tempfile.new('examples.rb')
     original_stdout = $stdout
-    $stdout = File.open(File::NULL, 'w')
-    begin
-      codeblock = read_examples
-      file.write(codeblock)
-      load(file.path)
-    ensure
-      $stdout = original_stdout
-      file.close
-      file.unlink
+    read_examples do |example, line_no|
+      _out, err = capture_io do
+        eval example
+      rescue => e
+        warn e
+      end
+      assert(
+        err.empty?,
+        "Example block doc/Examples.md:#{line_no} failed. See error below.\n#{err}"
+      )
     end
   end
 
@@ -36,14 +36,18 @@ class ExamplesTest < MiniTest::Test
   # read markdown line by line and if in a code block,
   # add line to string to be eval'd.
   def read_examples
-    code = ''
+    code = nil
+    line_no = nil
 
     File.open(examples_path, 'r') do |file|
       in_code_block = false
       file.each_line do |line|
         if !in_code_block && line.include?('```ruby')
+          code = ''
+          line_no = file.lineno
           in_code_block = true
         elsif in_code_block && line.include?('```')
+          yield(code, line_no)
           in_code_block = false
         elsif in_code_block
           code += line

```

A part from this tiny change, examples you made rocks, thank you !!

keithdoggett:
I think that makes more sense too and gives us more granularity, but in addition to the issue you pointed out there are 2 other issues.

1) After the factories section, I create one factory with `RGeo::Cartesian.factory` and assume it gets used for the rest of them. This would be easy enough to fix.
2) Sometimes I will reference a result from the previous example. For example, in the feature creation section, I do something like this for each geometry type:

```ruby
pt = factory.point(0, 0)
```

```ruby
pt2 = factory.parse_wkt("POINT (0 0)")
p pt == pt2
```
I guess I could remove those equality comparisons since that's the only place I reference previous results. What are your thoughts?

BuonOmo:
> I guess I could remove those equality comparisons since that's the only place I reference previous results. What are your thoughts?

Maybe the simplest approach could be to comment it out (the comment would say that they are equal)

Another more complete but complexe way to do it may be to add a comment _pre block_:

```markdown

<!--BEFORE-BLOCK
factory = ...
pt1 = ...
END-BEFORE-->

\```ruby
pt2 = ...
puts pt1 == pt2
\```
```

(the backslashes are just here to avoid mixing with github code block delimiters ^^

keithdoggett:
That makes sense. I'll just comment out those statements for now, but if we see the need to add them back later we can always add the comment for each block.

BuonOmo (approved):
LGTM!

Issue #231, opened 2020-11-16
Add a `geometry_collection#geometries` method


BuonOmo:
Nowadays, the process to retrieve all geometries from a geometry_collection is quite cumbersome. For instance, for a `MultiPolygon`:

```ruby
multi_polygon = RGeo::GeoJSON.decode(some_multi_polygon_geojson)

geometries = Array.new(object.num_geometries) { |i| multi_polygon.geometry_n(i) }
```

A more convenient API would be:

```ruby
multi_polygon = RGeo::GeoJSON.decode(some_multi_polygon_geojson)
geometries = multi_polygon.geometries
```

Going even further, we could include the ruby enumerable module, and add and `each` method that iterates over every single geometry, which would leverage all enumerable power and greatly simplify usage :)

PR #248, opened 2021-02-08
Add #geometries and utilize #each for Enumerable

mylesboone:
### Summary

This implements https://github.com/rgeo/rgeo/issues/231 .

### Other Information

There's a very basic test as well. More just a test to ensure that classes that include `BasicGeometryCollectionMethods` respond to enumerable methods.

### DIFF
--- a/lib/rgeo/impl_helper/basic_geometry_collection_methods.rb
+++ b/lib/rgeo/impl_helper/basic_geometry_collection_methods.rb
@@ -9,6 +9,8 @@
 module RGeo
   module ImplHelper # :nodoc:
     module BasicGeometryCollectionMethods # :nodoc:
+      include Enumerable
+
       attr_reader :elements
 
       def initialize(factory, elements)
@@ -35,6 +37,10 @@ module RGeo
 
       def each(&block)
         @elements.each(&block)
+      end
+
+      def geometries
+        @elements
       end
 
       def dimension
--- a/test/common/geometry_collection_tests.rb
+++ b/test/common/geometry_collection_tests.rb
@@ -45,6 +45,13 @@ module RGeo
           assert(@line1.eql?(geom[-1]))
         end
 
+        def test_enumerables
+          geom = @factory.collection([@point1, @line1])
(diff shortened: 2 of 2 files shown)


mylesboone:
FWIW -- I personally think that it'd be a lot cleaner to not access Enumerable methods directly on `BasicGeometryCollectionMethods`, but instead through `#goemetries`. In the same vein, removing `#each`. But, probably not worth the deprecation, etc. Up to you if you want to keep `Enumerable`.

BuonOmo (approved):
Thank you very much for your first contribution, and welcome ! 

The implementation LGTM 🙂 

About extending `geometries` instead of `BasicGeometryCollectionMethods`, do you really see a risk ? I've seen the latter in some gems already.

@keithdoggett I'll let you review as well, and I think we could include that in a minor pretty soon (documentation is waiting for the next update to be on rubydoc.info !)

mylesboone:
@BuonOmo I don't see a risk extending Enumerable in `BasicGeometryCollectionMethods`. I more just mean that we could take advantage of the fact that `geometries` is already an enumerable `Array` and for me, it would be more intuitive to call `my_geometry_collection.geometries.each{...}` over `my_geometry_collection.each{...}`, but I don't think there's any harm or risk in extending Enumerable in `BasicGeometryCollectionMethods`.

BuonOmo:
@mylesboone fine, in that case I'd say we keep both for developers lazyness. @keithdoggett any opinion ?

keithdoggett (approved):

keithdoggett:
@BuonOmo yeah let's just keep both. I don't see any reason to deprecate `each` right now.

@mylesboone looks good thanks for the PR!

mylesboone:
@keithdoggett not sure what you're wanting in the History.md -- looks like that documents release changes. I assume you'll want other changes before a minor version change?

keithdoggett:
@mylesboone sorry about the confusion. Usually there's a "HEAD" section with pending changes on master. I'll just merge and add that manually.

PR #249, opened 2021-02-09
Minor comestic tweaks to CI

mlarraz:
### Summary

* Better labels for tasks
* Use strings for Ruby versions to prevent rounding (3.0 -> 3, which would match 3.1)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,18 +12,19 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-          - ruby-head
-          - 3.0
-          - 2.7
-          - 2.6
-          - 2.5
+          - head
+          - '3.0'
+          - '2.7'
+          - '2.6'
+          - '2.5'
           - jruby
         os:
-          - ubuntu-latest
-          - windows-latest
-          - macos-latest
-    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


BuonOmo:
Great idea ! Would you have an idea to avoid the ❌ if mandatory checks passed as well ?

image

mlarraz:
Before and after:
image image

It's not an issue yet because `3.0` is the only release of `3.x` so far, but once `3.1` comes out `3` will resolve to that instead.

mlarraz:
> Great idea ! Would you have an idea to avoid the ❌ if mandatory checks passed as well ?

Unfortunately this is a UI issue I haven't found a good workaround for. The overall workflow itself still has a a success status, but this isn't clearly visible:
image

For now you could make certain checks required in the repo settings, to more clearly represent what is expected to pass.

BuonOmo:
Oh i see, the issue is only here when there is a `.0` version, nice catch 🙂

BuonOmo:
> For now you could make certain checks required in the repo settings, to more clearly represent what is expected to pass.

@keithdoggett I'll do so after we merge this PR

keithdoggett (approved):

keithdoggett:
Thanks again @mlarraz!

It's definitely annoying that the :x: shows up on builds that pass, especially since travis supported that functionality. I've seen some issues open for it in GitHub, so hopefully, it gets fixed soon.

BuonOmo (approved):
LGTM!

PR #250, opened 2021-02-09
Add Dependabot

mlarraz:
### Summary

Will help make sure dependencies are up to date

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,8 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


keithdoggett:
This looks good. Is it worthwhile to add one for our actions as well?

```
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      # Check for updates to GitHub Actions every weekday
      interval: "daily"
```

BuonOmo:
I am not in favor of that one: Dependabot add a lot of noise, and we're only two maintainers for now. Moreover, we only have dev dependencies so no bother for users of the gem. And GitHub security is on, for eventual CVEs we could encounter..

PR #251, opened 2021-03-30
Improve geos multipolygon creation performance

Quiwin:
Replace a manual validity check on MultiPolygon, which was iterating on
each polygon combination, by a direct call of geos isValid.

### Summary
Hello !

MultiPolygon creation was slower than expected on large [MultiPolygon](https://gist.github.com/Quiwin/d3b0f47d5669260fd46d8ea3ad47077a), it was due to an old check for validity which manually iterated on every polygon combination of the given multiPolygon.
Now directly call geos to check for validity.

### Other Information
Tested with above geojson:
```ruby
geojson = IO.read("france.geojson")
require 'micro_bench'
MicroBench.start;RGeo::GeoJSON.decode(geojson, geo_factory: RGeo::Cartesian.preferred_factory); puts "#{MicroBench.duration.round(2)} sec"
```
Before: 1.95 sec
After: 0.35 sec

I couldn't find the [exact geos issue](https://trac.osgeo.org/geos/search?milestone=on&changeset=on&ticket=on&wiki=on&q=multipolygon+valid&page=3&noquickjump=1) it tried to avoid
but I dont think there is much risk since the previous code written to circumvent it was more than 11 years ago https://github.com/rgeo/rgeo/commit/c44767eec656d9780919d007de41be8f19f773bc

Thanks!

### DIFF
--- a/ext/geos_c_impl/geometry_collection.c
+++ b/ext/geos_c_impl/geometry_collection.c
@@ -41,9 +41,6 @@ static VALUE create_geometry_collection(VALUE module, int type, VALUE factory, V
   VALUE cast_type;
   GEOSGeometry* geom;
   GEOSGeometry* collection;
-  char problem;
-  GEOSGeometry* igeom;
-  GEOSGeometry* jgeom;
 
   result = Qnil;
   Check_Type(array, T_ARRAY);
--- a/ext/geos_c_impl/geometry_collection.c
+++ b/ext/geos_c_impl/geometry_collection.c
@@ -89,10 +89,10 @@ static VALUE create_geometry_collection(VALUE module, int type, VALUE factory, V
     else {
       collection = GEOSGeom_createCollection_r(geos_context, type, geoms, len);
       // Check if MultiPolygon is valid.
-      if (collection &&
-          type == GEOS_MULTIPOLYGON
-          && (factory_data->flags & 1) == 0
-          && (GEOSisValid_r(geos_context, collection) == 0)) {
+      if (collection
+          && type == GEOS_MULTIPOLYGON
+          && !(factory_data->flags & 1)
+          && !GEOSisValid_r(geos_context, collection)) {
         GEOSGeom_destroy_r(geos_context, collection);
         collection = NULL;
       }
--- a/ext/geos_c_impl/geometry_collection.c
+++ b/ext/geos_c_impl/geometry_collection.c
@@ -88,31 +88,13 @@ static VALUE create_geometry_collection(VALUE module, int type, VALUE factory, V
     }
     else {
       collection = GEOSGeom_createCollection_r(geos_context, type, geoms, len);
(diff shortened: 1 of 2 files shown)


BuonOmo:
Looking at [history of the `IsValidOp.h`](https://github.com/libgeos/geos/commits/master/include/geos/operation/valid/IsValidOp.h) I don't see the point of that comment or code either.

@Quiwin knowing your sources, I guess your `france.geojson` polygon is [this one](https://github.com/gregoiredavid/france-geojson/blob/45daa2d/metropole.geojson).

I'm more in favor than not : it seems genuinely better.

@keithdoggett do you know the historical reason there, or do you think we should ping Daniel ? Also about the failing tests, we should work on that asap !

ext/geos_c_impl/geometry_collection.c
```diff
@@ -88,31 +88,13 @@ static VALUE create_geometry_collection(VALUE module, int type, VALUE factory, V
...
-        }
-        if (problem) {
-          GEOSGeom_destroy_r(geos_context, collection);
-          collection = NULL;
-        }
+      // Check if MultiPolygon is valid.
+      if (collection &&
+          type == GEOS_MULTIPOLYGON
+          && (factory_data->flags & 1) == 0
+          && (GEOSisValid_r(geos_context, collection) == 0)) {
```
BuonOmo, line 95: One small cosmetic change that feels clearer !
```suggestion
      if (collection
          && type == GEOS_MULTIPOLYGON
          && !(factory_data->flags & 1)
          && !GEOSisValid_r(geos_context, collection)) {
```

BuonOmo (approved):

keithdoggett:
Looks good! I'm not sure of the historical purpose of those extra checks. It seems that `GeosIsValid` supports all geometries, and this isn't used to identify a specific issue with validity since it doesn't do anything differently based on which check it fails.

@BuonOmo are all of the checks failing because of a floating-point comparison issue? I think I have a fix for that in the v3-dev branch we can probably cherry-pick.

BuonOmo:
As discussed, we should merge this one into the v3-dev to avoid having to backport of all @keithdoggett's work on floating point issues. @Quiwin could you rebase on this branch and we'll merge it there.

That will be one more great reason for using rgeo 3.0 :smile:

Quiwin:
> As discussed over the phone with @keithdoggett, we should merge this one into the v3-dev to avoid having to backport of all @keithdoggett's work on floating point issues. @Quiwin could you rebase on this branch and we'll merge it there.
> 
> That will be one more great reason for using rgeo 3.0 😄
> 
> And could you also add a note to the changelog ? (History.md file)

Done :)

BuonOmo:
Replace a manual validity check on MultiPolygon, which was iterating on
each polygon combination, by a direct call of `GEOSisValid`.

PR #253, opened 2021-04-15
Fix `Polygon#invalid_reason`

BuonOmo:
Fixes both the returned value, which should be `Qnil` under some
circumstances, and the related memory leak, pointer was not correctly
freed.

REPRO:

```ruby
#!/usr/bin/env ruby -I. -r./lib/rgeo.rb
require "get_process_mem"

pt = RGeo::Geos.factory.point(1,0)
rep = 10_000_000

10.times do
  rep.times { pt.invalid_reason }
  GC.start
  puts GetProcessMem.new.mb
end
```

@keithdoggett I'd like to rebase the v3-dev branch after merge if that is ok with you

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,8 @@
+### Unreleased
+
+* Fix a memory leak in CAPI's `#invalid_reason`
+* `#invalid_reason` now must return `nil` if polygon is valid.
+
 ### 2.2.0 / 2020-11-18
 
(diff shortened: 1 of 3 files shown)


keithdoggett (approved):
LGTM! Yes rebasing v3-dev to master sounds good.

BuonOmo:
@keithdoggett thanks, will do as soon as you bump the version for the release :)

PR #260, opened 2021-06-11
Update Enable-GEOS-and-Proj4-on-Heroku.md

earnold:
### Summary

This updates the documentation with what is necessary to get [rgeo-proj4](https://github.com/rgeo/rgeo-proj4) working on Heroku

See https://github.com/rgeo/rgeo-proj4/issues/17 for further details.

### DIFF
--- a/doc/Enable-GEOS-and-Proj4-on-Heroku.md
+++ b/doc/Enable-GEOS-and-Proj4-on-Heroku.md
@@ -9,7 +9,7 @@ If you need GEOS and PROJ extensions on Heroku, you need to install those librar
 ```bash
 echo 'libgeos-dev=3.7.1-1~pgdg18.04+1' > Aptfile # Use the version you want here
 echo 'libproj-dev' >> Aptfile                    # Same here, you can pin any version as well.
-echo 'proj-bin' >> Aptfile                       # Necessary if using rgeo-proj4 
+echo 'proj-bin' >> Aptfile                       # Necessary if using rgeo-proj4
 echo >> Aptfile
 git add Aptfile
 git commit -m 'Add Aptfile'
@@ -47,7 +47,7 @@ $ heroku buildpacks:add --index 1 heroku-community/apt
 
 If your Heroku environment has a config variable named `LD_LIBRARY_PATH`, please unset that.
 
-If you plan on using [rgeo-proj4](https://github.com/rgeo/rgeo-proj4) you must set the config variable named `PROJ_LIB` to `/app/.apt/usr/share/proj`
+If you plan on using [rgeo-proj4] you must set the config variable named `PROJ_LIB` to `/app/.apt/usr/share/proj`
 
 It is also really likely that you have the rgeo gem beforehand, in which case you will need to purge your application cache:
 
@@ -83,7 +83,7 @@ heroku plugins:install heroku-repo
 heroku repo:purge_cache
 ```
 
-**Note** This option will NOT work with [rgeo-proj4](https://github.com/rgeo/rgeo-proj4). 
+**Note** This option will NOT work with [rgeo-proj4].
 
 ## Check that geos is correctly installed
 
@@ -96,3 +96,4 @@ Note that checking `RGeo::Geos.supported?` might be misleading since this would 
 [heroku-buildpack-apt]: https://elements.heroku.com/buildpacks/heroku/heroku-buildpack-apt
 [heroku-buildpack-vendorbinaries]: https://github.com/diowa/heroku-buildpack-vendorbinaries
 [heroku-geo-buildpack]: https://github.com/heroku/heroku-geo-buildpack
+[rgeo-proj4]: https://github.com/rgeo/rgeo-proj4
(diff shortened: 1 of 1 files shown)


keithdoggett:
Awesome thanks for doing this! Going to link the original issue for reference (https://github.com/rgeo/rgeo-proj4/issues/17).

keithdoggett (approved):

doc/Enable-GEOS-and-Proj4-on-Heroku.md
```diff
@@ -79,6 +83,8 @@ heroku plugins:install heroku-repo
 heroku repo:purge_cache
 ```
 
+**Note** This option will NOT work with [rgeo-proj4](https://github.com/rgeo/rgeo-proj4).
```
BuonOmo, line 86: The buildpack's readme explicitely says that it works with gdal and proj for rgeo. Hence I think this should be an issue in the buildpack's repo (or a PR!).

Also one tiny change: if you could use the same way of adding links that the doc already use:

```suggestion
**Note** This option will NOT work with [rgeo-proj4]. 
```

and at the end of the doc:

```md
[rgeo-proj4]: https://github.com/rgeo/rgeo-proj4
```

BuonOmo (commented):
Thank you for improving that doc, and do not hesitate to do it again if you have something new!

The only issue I feel is left is the buildpack, which should just work. @keithdoggett do you know if the issue comes from our repo or the buildpack ?

doc/Enable-GEOS-and-Proj4-on-Heroku.md
keithdoggett: I think the issue is that with the update to v3 of RGeo-Proj4, it depends on PROJ 6.2+, but the buildpack installs PROJ 5.2 by default. It seems like there's a way to install a different version though. I don't use Heroku so I'm not familiar with the process, but that should get it to work.

earnold: @BuonOmo I update to using the style of links you suggested. It's neat! I didn't know markdown could do that.

earnold: @BuonOmo & @keithdoggett - I'll admit, after some initial difficulties with the buildpack I gave up on it. However, I totally _could_ work. I've updated the comment to reflect that.

doc/Enable-GEOS-and-Proj4-on-Heroku.md
```diff
@@ -79,6 +83,8 @@ heroku plugins:install heroku-repo
 heroku repo:purge_cache
 ```
 
+**Note** In order for this to work [rgeo-proj4] you will need to use ENV vars to update to PROJ 6.2+.
```
BuonOmo, line 86: @earnold thanks for the clarification. It looks like from the buildpack code (README is a bit unclear on that point) that the way to make it work is:

```bash
echo 6.2 > PROJ_VERSION
git add PROJ_VERSION
git commit -m "Set proj version for geo-buildpack"
```

Would you have an ENV on which you could test that this works ? We'll avoid someone opening another issue on that exact topic!

And if it doesn't work, one will have to post an issue to the buildpack repo (I may do it).

earnold: Happy to test - but might not get to it later this week / early next week.

BuonOmo: @earnold, that is fine for me. Let's say we merge asis in one month if it is stale. And someone will reopen an issue when they encounter the problem.

earnold: @BuonOmo - I gave the buildpack a brief test and I couldn't get it working. 

One nitpick: to set the heroku env var, the command to use is `heroku config:set PROJ_VERSION=6.2`

Even with that, however, I get an error: 
```
remote: -----> Building on the Heroku-20 stack
remote: -----> Using buildpacks:
remote:        1. https://github.com/heroku/heroku-geo-buildpack.git
remote:        2. heroku/ruby
remote: -----> Geo Packages (GDAL/GEOS/PROJ) app detected
remote: -----> Installing GDAL-2.4.0
remote: -----> Installing GEOS-3.7.2
remote: -----> Installing PROJ-6.2
remote:  !     Requested PROJ Version (6.2) is not available for this stack (heroku-20).
remote:  !     Aborting.
```

I also tried with PROJ_VERSION set to 6.3.1 and 7.2.1, and those failed as well. 

I can't rule out that it would be possible to get it working, but it's definitely not working for me.

BuonOmo: @earnold thank you for your time! I think we have a great reference now on why this is not working with the geo buildpack, and if one where to ask we could point toward that PR and tell the to open an issue or fix on the buildpack.

BuonOmo (approved):
I've offered one last tiny change reflecting the buildpack issues. Otherwise, LGTM!

keithdoggett:
Thanks @earnold!

PR #261, opened 2021-06-25
Fix core dump when cast returns a non Geos object

BuonOmo:
reproduction:

```bash
ruby -Ilib -rrgeo -rostruct -e '
   p RGeo::Geos.factory.point(1,1).contains?(OpenStruct.new(factory: RGeo::Geos.factory))
'
```


keithdoggett (approved):
LGTM! Nice catch.

PR #262, opened 2021-06-30
Add Segment Intersection Method and Sweepline Intersector Class

keithdoggett:
This is the first of three cherry-picked commits from the `valid-op` branch.

### Summary

* Adds a `segment_intersection` method to determine the intersection point between two segments.
* Adds a `SweeplineIntersector` class that can be used to determine the intersections between a list of segments.

### Other Information

I'm using a modified version of this [algorithm](https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm) for the `SweeplineIntersector`. Instead of using a BST for the status object, I'm just keeping all of the current segments in an array and comparing them all against each other. This means the worst case is O(n^2), but the average case is a lot better.

Below is the results of a test for finding the intersections in a convex linear ring with various number of vertices vs. the current `is_simple?` algorithm.

Current Algorithm:
```
       user     system      total        real
100 Vertices  0.002848   0.000000   0.002848 (  0.002845)
200 Vertices  0.011324   0.000000   0.011324 (  0.011344)
300 Vertices  0.025661   0.000173   0.025834 (  0.025841)
400 Vertices  0.045257   0.000000   0.045257 (  0.045267)
500 Vertices  0.070586   0.000000   0.070586 (  0.070602)
600 Vertices  0.101751   0.000000   0.101751 (  0.101772)
700 Vertices  0.137849   0.000000   0.137849 (  0.137879)
800 Vertices  0.177336   0.000000   0.177336 (  0.177373)
900 Vertices  0.225233   0.000000   0.225233 (  0.225284)
1000 Vertices  0.278194   0.000000   0.278194 (  0.278251)
2500 Vertices  1.741052   0.000000   1.741052 (  1.741438)
```

SweeplineIntersector:
```
       user     system      total        real
100 Vertices  0.002291   0.000000   0.002291 (  0.002304)
200 Vertices  0.004537   0.000169   0.004706 (  0.004704)
300 Vertices  0.007165   0.000114   0.007279 (  0.007281)
400 Vertices  0.009682   0.000000   0.009682 (  0.009687)
500 Vertices  0.012650   0.000000   0.012650 (  0.012669)
600 Vertices  0.015084   0.000000   0.015084 (  0.015090)
700 Vertices  0.017796   0.000000   0.017796 (  0.017803)
800 Vertices  0.020277   0.000000   0.020277 (  0.020283)
900 Vertices  0.022950   0.000000   0.022950 (  0.022955)
1000 Vertices  0.025088   0.000000   0.025088 (  0.025092)
2500 Vertices  0.067948   0.000000   0.067948 (  0.067964)
```

### Future Work

Eventually, we may want to modify the `segment_intersection` method to return some type of struct that indicates the intersection type as well. Colinear intersections just return one point from the intersection, but in the future, we may want to get the subsegment from the intersection. This could cause issues with validity checking in a linear ring like this: `LINESTRING(0 0, 1 0, 1 1, 0.5 1, 0.5 1.5, 0.5 1, 0 1, 0 0)`
image

Where the segments in the spike might not be included in the `proper_intersections` array, but don't intersect in a valid way.


lib/rgeo/cartesian/calculations.rb
```diff
@@ -69,47 +69,236 @@ def contains_point?(p)
...
+      #
+      # Compares the new segment from the +is_start+ event to each observed segment
+      # then adds it to +observed_segments+. Records any intersections in to the
+      # returned array.
+      #
+      # @return [Array<RGeo::Cartesian::SweeplineIntersector::Intersection>]
+      def intersections
+        return @intersections if @intersections.size.positive?
+
+        observed_segments = []
```
BuonOmo, line 213: I guess this could be a `Set`, as fast for iteration, but way faster for deletion

lib/rgeo/cartesian/calculations.rb
```diff
@@ -69,47 +69,236 @@ def contains_point?(p)
...
+      # +observed_segments+ array while their ending event has not been popped
+      # from the queue.
+      #
+      # Compares the new segment from the +is_start+ event to each observed segment
+      # then adds it to +observed_segments+. Records any intersections in to the
+      # returned array.
+      #
+      # @return [Array<RGeo::Cartesian::SweeplineIntersector::Intersection>]
+      def intersections
+        return @intersections if @intersections.size.positive?
```
BuonOmo, line 211: rather than initializing and checking for the size, you could check for definition. It is a bit faster, and IMHO clearer on intention:

```suggestion
        return @intersections if defined?(@intersections)
```

lib/rgeo/cartesian/calculations.rb
```diff
@@ -69,47 +69,236 @@ def contains_point?(p)
...
       end
     end
+
+    # Implements a Sweepline intersector to find all intersections
+    # in a group of segments. The idea is to use a horizontal line starting
+    # at y = +Infinity that sweeps down to y = -Infinity and every time it hits
+    # a new line, it will check if it intersects with any of the segments
+    # the line currently intersects at that y value.
+    # This is a more simplistic implementation that uses an array to hold
+    # observed segments instead of a sorted BST, so performance may be significantly
```
BuonOmo, line 158: I am a bit unsure of where it would be more appropriate, could you indicate me ?

BuonOmo (commented):
Really cool, thank you for this implementation!

I feel like it could be in a separated file. But that is just IMHO.

I've made a few reviews, but everything feels great globally, it is just about tiny fixes.


On the giving segments versus points, I think even GEOS gives only points, hence I'm ensure we need to go that far!

lib/rgeo/cartesian/calculations.rb
```diff
@@ -69,47 +69,236 @@ def contains_point?(p)
...
+      # Compares the new segment from the +is_start+ event to each observed segment
+      # then adds it to +observed_segments+. Records any intersections in to the
+      # returned array.
+      #
+      # @return [Array<RGeo::Cartesian::SweeplineIntersector::Intersection>]
+      def intersections
+        return @intersections if @intersections.size.positive?
+
+        observed_segments = []
+        eq = Queue.new
```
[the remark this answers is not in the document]
keithdoggett: Yeah, I saw that but you're probably right it's easy enough to just pop from an array.

lib/rgeo/cartesian/calculations.rb
```diff
@@ -69,47 +69,236 @@ def contains_point?(p)
...
+          s_event.is_start = false
+          e_event.is_start = true
+        elsif s.x < e.x
+          s_event.is_start = true
+          e_event.is_start = false
+        elsif s.x > e.x
+          s_event.is_start = false
+          e_event.is_start = true
+        else
+          # degenerate case, should probably never happen
```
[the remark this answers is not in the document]
keithdoggett: I think I'll just convert the last elsif to an else. If we raise here or don't label the events we'd have to add a check for that upstream.

lib/rgeo/cartesian/calculations.rb
keithdoggett: image
```
LINESTRING(1 1, 0 1, 0 0, 0.125 0.5, 0.25 0, 0.375 0.5, 0.5 0, 0.625 0.5, 0.75 0, 0.875 0.5, 1 0, 1 1)
```

In this case, since we are comparing all `observed_segments` against each other, this would be about O(n^2) since pretty much every segment is being observed at once here. If we were using a BST instead of an array or set, we would sort it so that only adjacent segments are compared to each other, but it's a lot more work to implement the tree and keep track of switches in order. 

In this case,

image

```
LINESTRING (2.0 0.0, 1.847759065 -0.7653668647, 1.4142135624 -1.4142135624, 0.7653668647 -1.847759065, 0.0 -2.0, -0.7653668647 -1.847759065, -1.4142135624 -1.4142135624, -1.847759065 -0.7653668647, -2.0 0.0, -1.847759065 0.7653668647, -1.4142135624 1.4142135624, -0.7653668647 1.847759065, 0.0 2.0, 0.7653668647 1.847759065, 1.4142135624 1.4142135624, 1.847759065 0.7653668647, 2.0 0.0)
```

There are only ever 2 `observed_segments` at once so it will run closer to O(nlogn).

keithdoggett: There's also another strategy that GEOS uses where they break the segments up into monotone pieces and run the intersection on each piece then combine the results. 

I don't think it's really necessary we do either of these. If performance is that big of a concern you should use GEOS instead of the simpler Ruby implementation.

lib/rgeo/cartesian/calculations.rb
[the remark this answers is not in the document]
keithdoggett: I ended up condensing the flow to an if/else.

lib/rgeo/cartesian/calculations.rb
BuonOmo: @keithdoggett I definitely agree, thanks for the pointer though!

lib/rgeo/cartesian/sweepline_intersector.rb
```diff
@@ -0,0 +1,149 @@
...
+            observed_segments.each do |oseg|
+              int_pt = seg.segment_intersection(oseg)
+              if int_pt
+                intersect = Intersection.new(int_pt, seg, oseg)
+                @intersections << intersect
+              end
+            end
+            observed_segments << seg
+          else
+            observed_segments.delete_if { |oseg| oseg == seg }
```
BuonOmo, line 85: Using `#delete_if` runs a `#select` against the keys, making the usage of a set inefficient. `#delete` is of choice here!
```ruby
require "benchmark/ips"
require "set"

set = Array.new(10_000) { rand }.to_set

Benchmark.ips do |x|
  x.report("#delete_if") { r = rand;set.delete_if { _1 == r} }
  x.report("#delete") { r = rand;set.delete(r) }
  x.compare!
end
```

```
          #delete_if   114.000  i/100ms
             #delete   727.950k i/100ms
          #delete_if      1.136k (± 0.7%) i/s -      5.700k in   5.019563s
             #delete      7.224M (± 1.3%) i/s -     36.398M in   5.039573s

Comparison:
             #delete:  7223549.1 i/s
          #delete_if:     1135.6 i/s - 6360.90x  (± 0.00) slower
```

BuonOmo (approved):
One last review, otherwise I think we're ready to ship, thanks!

lib/rgeo/cartesian/sweepline_intersector.rb
keithdoggett: Thanks! Good catch.

Issue #247, opened 2021-02-04
RGeo 3.0 Roadmap


BuonOmo:
## Summary

We want to simplify the overall interface, being more _ruby-stylish_ (a.k.a simple). Refactor code to improve maintainability. Here are the main axis on which we want to focus for the next major:

- Loose Invalidity handling for geometries
- CAPI Refactoring
- Better GeoJSON integration
- Consistency and clarification of interface

## <a id=loose></a>Loose Invalidity handling for geometries

This is currently the most critical item. We want to always have loose invalidity handling and never raise on creation, which will allow us to get rid of `uses_lenient_assertions`. This will have the benefit of **always raise when a method cannot be use** while not making usage cumbersome. For instance, one may want to parse a polygon even if it is not simple, however, we should not give him a bad area computation because it was not verified.

Work on that topic has already been started in the https://github.com/rgeo/rgeo/tree/fix/correct-area-when-invalid-polygon branch. However, there are still performance issues that should be investigated.

- [x] https://github.com/rgeo/rgeo/issues/256.
  - https://github.com/rgeo/rgeo/pull/262 
  - https://github.com/rgeo/rgeo/pull/271

## <a id=capi></a>CAPI Refactoring

- [x] C Extension Update. Change data wrappers at least.
  - https://github.com/rgeo/rgeo/issues/273
  - change notice and error handlers (https://github.com/rgeo/rgeo/pull/286)
  - see if we really need the multithreaded api (or if ruby GIL is enough protection)
- [ ] https://github.com/rgeo/rgeo/issues/272

## <a id="geojson"></a>Better GeoJSON integration

GeoJSON is one of the most used way to parse data into `RGeo::Geometry` object, yet it is cumbersome to use.

- [x] https://github.com/rgeo/rgeo-geojson/issues/54
- [x] #266
- [x] https://github.com/rgeo/rgeo-geojson/pull/53 (we should not have to do such thing as https://github.com/rgeo/rgeo-geojson/issues/34#issuecomment-721990270)
- [ ] https://github.com/rgeo/rgeo-geojson/issues/55

## Consistency and clarification of interface

- [x] #267
- [x] #264
- [x] We should have a simple way to iterate through a collection rather than `geometry_n` (https://github.com/rgeo/rgeo/issues/231)
- [x] #265

The last item here is to be discussed: we do not want to add something that may make the interface more complex. If you read until here, Please mark your interest for this feature. It may be the case that this is only useful for us maintainers when creating examples and tests.

[...]

PR #268, opened 2021-07-07
Changed all is_*? method names to *?

### LINKED ISSUE #264: `is_empty?` should be replaced with `empty?` (Or at least alias it for a smoother transition). This goes for every `is_*?` method.

stephenandersondev:
### Summary

Changed all is_*? method names to *? on the following:

- is_empty? to empty?
- is_simple? to simple?
- is_closed? to closed?
- is_point? to point?
- is_ring? to ring?
- is_capi_geos? to capi_geos?
- is_ffi_geos? to ffi_geos?
- is_geos? to goes?

Closes #264

### DIFF
--- a/History.md
+++ b/History.md
@@ -161,7 +161,7 @@ Fixes
 
 ### 0.3.15 / 2012-08-02
 
-*   The class shuffle in 0.3.14 broke RGeo::Geos.is_geos? and similar. Fixed.
+*   The class shuffle in 0.3.14 broke RGeo::Geos.geos? and similar. Fixed.
 
 
 ### 0.3.14 / 2012-07-08
@@ -692,7 +692,7 @@ planned is the 0.2 alpha release.
     the final name for this function.
 *   Some new tests and a lot of fixes in SimpleMercator and SimpleSpherical.
 *   Implemented a few more pieces of SimpleSpherical. Notably,
-    LineString#is_simple? (which should now allow LinearRing to work).
+    LineString#simple? (which should now allow LinearRing to work).
 *   Classes that included Features::Geometry had their === operator
     erroneously overridden. Fixed.
 *   A few more documentation updates.
--- a/History.md
+++ b/History.md
@@ -161,7 +161,7 @@ Fixes
 
 ### 0.3.15 / 2012-08-02
 
-*   The class shuffle in 0.3.14 broke RGeo::Geos.geos? and similar. Fixed.
+*   The class shuffle in 0.3.14 broke RGeo::Geos.is_geos? and similar. Fixed.
 
 
 ### 0.3.14 / 2012-07-08
@@ -692,7 +692,7 @@ planned is the 0.2 alpha release.
     the final name for this function.
 *   Some new tests and a lot of fixes in SimpleMercator and SimpleSpherical.
 *   Implemented a few more pieces of SimpleSpherical. Notably,
-    LineString#simple? (which should now allow LinearRing to work).
+    LineString#is_simple? (which should now allow LinearRing to work).
 *   Classes that included Features::Geometry had their === operator
     erroneously overridden. Fixed.
 *   A few more documentation updates.
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -1095,8 +1095,8 @@ void rgeo_init_geos_geometry(RGeo_Globals* globals)
   rb_define_method(geos_geometry_methods, "boundary", method_geometry_boundary, 0);
   rb_define_method(geos_geometry_methods, "_as_text", method_geometry_as_text, 0);
   rb_define_method(geos_geometry_methods, "as_binary", method_geometry_as_binary, 0);
-  rb_define_method(geos_geometry_methods, "is_empty?", method_geometry_is_empty, 0);
-  rb_define_method(geos_geometry_methods, "is_simple?", method_geometry_is_simple, 0);
+  rb_define_method(geos_geometry_methods, "empty?", method_geometry_is_empty, 0);
+  rb_define_method(geos_geometry_methods, "simple?", method_geometry_is_simple, 0);
   rb_define_method(geos_geometry_methods, "equals?", method_geometry_equals, 1);
   rb_define_method(geos_geometry_methods, "==", method_geometry_equals, 1);
   rb_define_method(geos_geometry_methods, "rep_equals?", method_geometry_eql, 1);
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -1096,9 +1096,7 @@ void rgeo_init_geos_geometry(RGeo_Globals* globals)
   rb_define_method(geos_geometry_methods, "_as_text", method_geometry_as_text, 0);
   rb_define_method(geos_geometry_methods, "as_binary", method_geometry_as_binary, 0);
   rb_define_method(geos_geometry_methods, "empty?", method_geometry_is_empty, 0);
-  rb_define_method(geos_geometry_methods, "is_empty?", method_geometry_is_empty, 0);
   rb_define_method(geos_geometry_methods, "simple?", method_geometry_is_simple, 0);
-  rb_define_method(geos_geometry_methods, "is_simple?", method_geometry_is_simple, 0);
   rb_define_method(geos_geometry_methods, "equals?", method_geometry_equals, 1);
   rb_define_method(geos_geometry_methods, "==", method_geometry_equals, 1);
   rb_define_method(geos_geometry_methods, "rep_equals?", method_geometry_eql, 1);
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -1096,7 +1096,9 @@ void rgeo_init_geos_geometry(RGeo_Globals* globals)
   rb_define_method(geos_geometry_methods, "_as_text", method_geometry_as_text, 0);
   rb_define_method(geos_geometry_methods, "as_binary", method_geometry_as_binary, 0);
   rb_define_method(geos_geometry_methods, "empty?", method_geometry_is_empty, 0);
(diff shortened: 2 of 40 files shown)


stephenandersondev:
This is my first open-source contribution, so please let me know if I've missed any steps.

keithdoggett:
@stephenandersondev thanks for submitting this I think this looks good, but I'll take a deeper look soon. @BuonOmo do you think we should alias all the methods to accept the is_*? form as well? It would be good for backwards compatibility.

Also @stephenandersondev this is going to be part of our V3 release, so this needs to be merged into the `v3-dev` branch.

stephenandersondev:
@keithdoggett Just changed the branch to `v3-dev`, thanks for the heads up!

BuonOmo:
> This is my first open-source contribution ever, so please let me know if I've missed any steps. Thanks!

Thank you for doing it there!

---

@keithdoggett @stephenandersondev I thought a bit about it, and I think we can do some easier transition for users, that would even allow us to merge this in v2.

I'm thinking about keep `is_*` methods as well for a transition period, yet printing a deprecation warning. For instance:

```ruby
def is_simple?
	warn "is_*? methods are deprecated, please use the *? counterpart, will be removed in v3" unless ENV["RGEO_SILENCE_DEPRECATION"]

	simple?
end
```

Of course, this might also be done using metaprogramming (for instance with `method_missing`), yet I'm not sure this is necessary.


What do you think about that idea, and @stephenandersondev could you implement it in this PR? 🙂

stephenandersondev:
@BuonOmo Yes I can definitely implement this PR, I will have it ready shortly 👍

stephenandersondev:
@BuonOmo I got the alias methods put in, have a couple of questions though:

- I didn't see a branch that referenced v2 so I assumed it was `master` since it's the current version? Let me know if otherwise and I can change it.
- I wasn't sure if we needed to add the alias methods to the tests? If so, let me know what you think that should look like and I'll get it in.

ext/geos_c_impl/geometry.c
```diff
@@ -1095,8 +1095,8 @@ void rgeo_init_geos_geometry(RGeo_Globals* globals)
   rb_define_method(geos_geometry_methods, "boundary", method_geometry_boundary, 0);
   rb_define_method(geos_geometry_methods, "_as_text", method_geometry_as_text, 0);
   rb_define_method(geos_geometry_methods, "as_binary", method_geometry_as_binary, 0);
-  rb_define_method(geos_geometry_methods, "is_empty?", method_geometry_is_empty, 0);
-  rb_define_method(geos_geometry_methods, "is_simple?", method_geometry_is_simple, 0);
+  rb_define_method(geos_geometry_methods, "empty?", method_geometry_is_empty, 0);
+  rb_define_method(geos_geometry_methods, "simple?", method_geometry_is_simple, 0);
```
BuonOmo, line 1099: The geos methods do not implement `Feature` base methods in v2 unfortunately.

Hence we also need to create the deprecation warning. You can do that one in ruby, in `lib/rgeo/geos/capi_feature_classes.rb` :)

ext/geos_c_impl/geometry_collection.c
```diff
@@ -623,7 +623,7 @@ void rgeo_init_geos_geometry_collection(RGeo_Globals* globals)
   geos_multi_line_string_methods = rb_define_module_under(globals->geos_module, "CAPIMultiLineStringMethods");
   rb_define_method(geos_multi_line_string_methods, "geometry_type", method_multi_line_string_geometry_type, 0);
   rb_define_method(geos_multi_line_string_methods, "length", method_multi_line_string_length, 0);
-  rb_define_method(geos_multi_line_string_methods, "is_closed?", method_multi_line_string_is_closed, 0);
+  rb_define_method(geos_multi_line_string_methods, "closed?", method_multi_line_string_is_closed, 0);
```
BuonOmo, line 626: Same here unfortunately

ext/geos_c_impl/line_string.c
```diff
@@ -663,8 +663,8 @@ void rgeo_init_geos_line_string(RGeo_Globals* globals)
   rb_define_method(geos_line_string_methods, "end_point", method_line_string_end_point, 0);
   rb_define_method(geos_line_string_methods, "project_point", method_line_string_project_point, 1);
   rb_define_method(geos_line_string_methods, "interpolate_point", method_line_string_interpolate_point, 1);
-  rb_define_method(geos_line_string_methods, "is_closed?", method_line_string_is_closed, 0);
-  rb_define_method(geos_line_string_methods, "is_ring?", method_line_string_is_ring, 0);
+  rb_define_method(geos_line_string_methods, "closed?", method_line_string_is_closed, 0);
+  rb_define_method(geos_line_string_methods, "ring?", method_line_string_is_ring, 0);
```
BuonOmo, line 667: and there

BuonOmo (commented):
@stephenandersondev thank you for the addition.

> I didn't see a branch that referenced v2 so I assumed it was master since it's the current version? Let me know if otherwise and I can change it.

You're right, v2 is the current version, we're using the master branch.

> I wasn't sure if we needed to add the alias methods to the tests? If so, let me know what you think that should look like and I'll get it in.

IMHo, that is not needed.

---

I've one last review concerning the geos capi, unfortunately we need alias for those methods as well. I've adressed which methods and where they should be aliased. I can do it and push to your branch later if you don't have more time 🙂 (I'll let you decide though, your branch, your ownership!)

After that I think we'll be ready to ship.

stephenandersondev:
So sorry about that @BuonOmo. I'm admittedly very new to RGeo and this is my first time ever looking at C code 😅  I think I got the aliasing right? Although, you may mean to alias do the aliasing on the ruby side which is where I would add the warning back in at? If this is wrong (which I feel it is) would you be able to do one as an example?  I would be happy to do the rest 🙂  Sorry again, still learning this stuff.

BuonOmo:
@stephenandersondev no problem, I'll give you some pointers.

Let's try with the `empty?` GEOS method for instance. The idea will be to implement `is_empty?` on the ruby side, to have something simpler to code. Performances will not be an issue anyway for such a trivial task (proxying to the actual `empty?` method).

We first need to know where the method is implemented. The line below gives us this information:

```c
rb_define_method(geos_geometry_methods, "empty?", method_geometry_is_empty, 0);
```

<sup>
[`rb_define_method` reference](https://github.com/ruby/ruby/blob/HEAD/doc/extension.rdoc#method-and-singleton-method-definition-)
</sup>

We can see it is defined under `geos_geometry_methods`. And this correspond to the ruby module `XX::CAPIGeometryMethods` according to:

```c
geos_geometry_methods = rb_define_module_under(globals->geos_module, "CAPIGeometryMethods");
```

This module can be found when searching the project, and you'll most likely end up in `lib/rgeo/geos/capi_feature_classes.rb`, you now know you can define the method here:

```ruby
module RGeo
  module Geos
    module CAPIGeometryMethods # :nodoc:
      include Feature::Instance

      def is_empty?
        warn "support..."
        empty?
      end
...
```

Is that clear enough? If not, I'll just takeover this last bit and you'll see my commit as an example 🙂

stephenandersondev:
@BuonOmo Thank you so much for taking the time to clarify this for me, I learned something new because of it 😄  I believe this last commit I pushed does the trick. Please let me know if otherwise.  Thanks again for your help!

BuonOmo (approved):
LGTM!

BuonOmo:
@keithdoggett do you want to process to one last review?

BuonOmo:
Oh, my bad, rubocop is not passing anymore, @stephenandersondev the trick is to disable the `Naming/PredicateName` cop (see https://docs.rubocop.org/rubocop/1.3/configuration.html#disabling-cops-within-source-code)

stephenandersondev:
So sorry about that 🤦🏻  Just committed the disabling of it. Thank you so much for your patience and guidance on this!

BuonOmo:
@stephenandersondev sorry I should have been a bit clearer.... I think that this should be inlined:

```ruby
def is_simple? # rubocop:disable Naming/PredicateName
```

This will ensure that when getting rid of every `is_` method, we'll also get rid of that comment. Moreover, it tells a ready why this comment is here.

stephenandersondev:
@BuonOmo Oops sorry about that, they're inline now 👍🏻

History.md
```diff
@@ -161,7 +161,7 @@ Fixes
 
 ### 0.3.15 / 2012-08-02
 
-*   The class shuffle in 0.3.14 broke RGeo::Geos.is_geos? and similar. Fixed.
+*   The class shuffle in 0.3.14 broke RGeo::Geos.geos? and similar. Fixed.
```
keithdoggett, line 164: Not sure we should change this since it's describing a past change.

History.md
```diff
@@ -692,7 +692,7 @@ planned is the 0.2 alpha release.
     the final name for this function.
 *   Some new tests and a lot of fixes in SimpleMercator and SimpleSpherical.
 *   Implemented a few more pieces of SimpleSpherical. Notably,
-    LineString#is_simple? (which should now allow LinearRing to work).
+    LineString#simple? (which should now allow LinearRing to work).
```
keithdoggett, line 695: Not sure we should change this since it's describing a past change.

keithdoggett (commented):
Thanks for the work @stephenandersondev and @BuonOmo! These look good for the most part. My only suggestion is that we don't modify the `History.md` file because that's supposed to reflect historical changes that happened and probably shouldn't be modified only appended to.

stephenandersondev:
@keithdoggett Ahhh that's a great point. I just reverted them back. Thanks for the head up!

keithdoggett (approved):
Awesome, thanks @stephenandersondev! LGTM

@BuonOmo should be good to merge as is unless you want to squash the commits first.

BuonOmo:
Merged! Thanks again @stephenandersondev  :)

stephenandersondev:
Thank you both! I look forward to working with you again :)

BuonOmo:
well if you see something on which you could work on the [roadmap](https://github.com/rgeo/rgeo/issues/247), go ahead 🙂

PR #269, opened 2021-07-07
Add Planar Graph

keithdoggett:
### Summary

Adds a PlanarGraph implementation based on a Doubly Connected Edge List (DCEL). A DCEL is a graph data structure of
vertices and half-edges. Half-edges are directed edges where one segment can be represented by a pair of half-edges.
Half-edges are linked so that faces can be traversed (although faces are not stored).

The PlanarGraph accepts an array of segments to build it and additional edges can be added via the add_edge and add_edges methods. When edges are added, intersections are automatically recomputed and handled. When possible, existing half-edges are preserved so that geometries can reference those half-edges.

Also adds GeometryGraph which is a convenient way to interface geometries to a planar graph and provides ways to find important half-edges in each geometry.

[Algorithm](https://en.wikipedia.org/wiki/Doubly_connected_edge_list). I can provide more info about it if needed. It's not too complex, just a lot of bookkeeping.

### Purpose

- A few validation checks become fairly trivial with the use of a graph (disconnected interior and consistent area, for example).
- Later on, this could be extended slightly to compute the relationships between geometries (intersection, difference, etc.).

### Notes

GEOS does something slightly different where their edges are an entire linestring, not just a single edge. This reduces the size and can help simplify computing relationships between geometries, but updating and finding split edges becomes more tedious.


lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
+# frozen_string_literal: true
+
+module RGeo
+  module Cartesian
+    module Graphs
```
keithdoggett, line 5: Maybe a different name?

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+          hedges << self
+
+          n = self.next
+          until n.eql?(self) || n.nil?
+            yield(n) if block_given?
+            hedges << n
+            n = n.next
+          end
+          hedges
+        end
```
keithdoggett, line 66: This might not be the best name for the method.

Also, it could be possible that this ends up in an infinite somehow. I could add a counter and if the counter exceeds the number of vertices or edges, then it should return what it is has or raise an error.

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+
+        def initialize(origin)
+          @origin = origin
+          @twin = nil
+          @next = nil
+          @prev = nil
+        end
+        attr_reader :origin
+        attr_accessor :twin, :next, :prev
```
keithdoggett, line 41: Perhaps not the smartest to use `next` since that's a keyword. I would use `n` and `p` but `p` is also a keyword.

BuonOmo: Next is okay since it is a keyword and we are most likely to call it in a context that has no ambiguity. Since p is a method it would be worse. 

A solution though if you prefer : `next_hedge`

lib/rgeo/cartesian/planar_graph.rb
BuonOmo: Maybe I am overlooking something, but couldn't hedges be a set, and we would check for already existing hedge in this set each time ?

Also, why do you feel like each is not a good name ? 

One issue I could think of is that we're calling it on an edge, yet it returns the whole graph containing that edge. If that is correct, could it be `graph` or `related_hedges` ?

lib/rgeo/cartesian/planar_graph.rb
BuonOmo: Why not `PlanarGraph` (as suggested by the file). Other classes could be subclasses. Except for `GeometryGraph` IMHO

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+        # Creates 2 half edges from an edge.
+        # They will be assigned as each other's twins.
+        # The Half Edges will be returned in the order of points
+        # of the edge (start, end).
+        #
+        # @param edge [RGeo::Cartesian::Segment]
+        #
+        # @return [Array]
+        def self.from_edge(edge)
+          # TODO: should we raise an error if the edge is degenerate?
```
BuonOmo, line 25: I guess so :)

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+            n = n.next
+          end
+          hedges
+        end
+
+        # Return the destination of the half edge
+        #
+        # @return [RGeo::Feature::Point]
+        def destination
+          twin.origin
```
BuonOmo, line 72:
```suggestion
          twin&.origin
```

This will help you later, for instance in the `#to_s` method

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+      #
+      # Upon instantiation, the graph will compute the intersections using
+      # the SweeplineIntersector, populate the incident_edges map, and
+      # link all cyclic edges.
+      class PlanarGraph
+        # Create a new PlanarGraph
+        #
+        # @param edges [Array<RGeo::Cartesian::Segment>] of Segments
+        def initialize(edges = nil)
+          edges = [] if edges.nil?
```
BuonOmo, line 109: why not
```suggestion
        def initialize(edges = [])
```

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+          @edges = []
+          @incident_edges = {}
+
+          # this could be less efficient than computing the splits, then
+          # adding adding them and linking them since this ends up creating
```
BuonOmo, line 114:
```suggestion
          # adding them and linking them since this ends up creating
```

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+          # half edges that are potentially immediately split,
+          # but this creates a more consistent flow for adding edges.
+          create_half_edges
```
BuonOmo, line 117: Really not sure, seems like this call is useless, since `@edges` is an empty array

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+
+              hedges[-1].prev = hedges[0].twin
+              hedges[0].twin.next = hedges[-1]
+
+            # handle case of dangling line
+            else
+              e = hedges[0]
+              e.twin.next = e
+              e.prev = e.twin
+            end
```
BuonOmo, line 236: It seems like it is the same as:
```suggestion
            # Note that for a dangling line (single edge), first edge will be its own twin.
            [*hedges, hedges.first].each_cons(2) do |e1, e2|
              e1.prev = e2.twin
              e2.twin.next = e1
            end
```

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+          end
+        end
+
+        # It is possible that intersections occur when new edges are added.
+        # This will split those edges into more half-edges while preserving
+        # the existing half-edges when possible since geometries may reference
+        # them.
```
BuonOmo, line 243: Not entirely sure I get the underlying issue there..

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+
+        # Adds a geometry to the graph and finds its
+        # reference HalfEdge(s).
+        #
+        # @param geom [RGeo::Feature::Instance]
+        def add_geometry(geom)
+          case geom
+          when Feature::Point
+            # Can't handle points yet, so just add an empty entry for them
+            @geom_edges << GeomEdge.new
```
BuonOmo, line 302: I guess we'll see in the next PR that we don't use those for points, right?

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+          when Feature::LineString, Feature::LinearRing
+            add_line_string(geom)
+          when Feature::Polygon
+            add_polygon(geom)
+          when Feature::GeometryCollection
+            add_collection(geom)
+          end
```
BuonOmo, line 309: shouldn't we raise otherwise?

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+        end
+
+        # Adds a LineString or LinearRing
+        # to the graph.
+        #
+        # @param geom [RGeo::Feature::LineString]
+        def add_line_string(geom)
+          # TODO: should we handle empty linestrings?
```
BuonOmo, line 317: can we handle those smoothly, or do you talk about raising ?

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+        end
+
+        # Adds a Polygon to the graph.
+        #
+        # Will add each shell seperately and find a CCW half-edge
+        # for the exterior ring and a CW half-edge for interior rings
+        # since these are designated as being on the interior of the polygon.
+        #
+        # It is possible that the half-edge will be nil if a suitable one cannot
+        # be found. This indicates the polygon is likely invalid.
```
BuonOmo, line 334: _likely_ here is a bit scary. Is there a possibility it is valid ?

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+            pts = []
+            hedge.each do |he|
+              pts << he.origin
+            end
+            pts << hedge.origin
+
+            lr = parent_geometry.factory.line_string(pts)
+            return hedge if Analysis.ring_direction(lr) == ccw_target
+          end
+          nil
```
BuonOmo, line 395:
```suggestion
          colinear_hedges.find do |hedge|
            pts = []
            hedge.each do |he|
              pts << he.origin
            end
            pts << hedge.origin

            lr = parent_geometry.factory.line_string(pts)
            Analysis.ring_direction(lr) == ccw_target
          end
```

sorry, this is just me trying to be fancy ^^

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,400 @@
...
+          # segment of the ring.
+          start_seg = Segment.new(ring.start_point, ring.point_n(1))
+          end_seg = Segment.new(ring.point_n(ring.num_points - 2), ring.end_point)
+          colinear_hedges = hedges.select do |he|
+            start_seg.side(he.destination).zero? || end_seg.side(he.destination).zero?
+          end
+
+          colinear_hedges.each do |hedge|
+            pts = []
+            hedge.each do |he|
```
BuonOmo, line 387: Here comes the cycling issue you were talking about then.

I guess it is important not to get stuck in a infinite loop. Either the Set will be enough, or we need to find something else.

And, you're right, `each` doesn't feel correct there.

BuonOmo (commented):
Thank you for that complete implementation, at this point I'm just waiting for the third PR to see usage of all that!

I've already made a few comments, tried to answer you questions, and ask some questions of my own. I think I'll have to review that completely a second time just to be sure I've fully understood it.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: > Maybe I am overlooking something, but couldn't hedges be a set, and we would check for already existing hedge in this set each time ?

Yes, it probably could be, but would it be weird to return a set? I guess it doesn't matter since a set is still enumerable.

> Also, why do you feel like each is not a good name ?

> One issue I could think of is that we're calling it on an edge, yet it returns the whole graph containing that edge. If that is correct, could it be graph or related_hedges ?

That's exactly why I thought `each` might not make sense. It's not like `HalfEdge` is a collection, so calling each on it and returning a bunch of hedges seemed weird since `each` is generally reserved for collections/enumerables. I was thinking `cycle`, but `related_hedges` might be clearer.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: That's a good point. I'll likely leave it as `next` then.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: I guess my concern with raising an error is that the context this will be created in is when we're checking for validity (possibly more later). So we'd have to make sure that it is handled properly upstream. I'll think about it more, but it probably does make sense to raise and we'll just need to keep track of all the errors.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: I know JavaScript you're not really supposed to use objects as a default value since if they're mutated, the function may still reference that object. Not sure if it's the case with Ruby as well.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: If we pass in `edges` it won't be. But for a `GeometryGraph` it will be.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: Say we are making a `GeometryGraph` with the following geometry `GEOMETRYCOLLECTION(POLYGON((0 0,0 1,1 1,1 0,0 0)),LINESTRING(0.5 0.5,0.5 -0.5))`.
image

We add each component of the collection separately, so first we will add the polygon and create the half-edges (only showing the bottom).

image

The `GeometryGraph` will reference the half-edge on the interior of the polygon in its `geom_edges`. This is how it can access a half-edge on the interior of the polygon for future uses.

Now, when we go to overlay the next component, we need to split those half-edges since there's an intersection.
image

But it's important that we don't just remove the existing half-edges and replace them (although that would be easier), since the `GeometryGraph` will hold a reference to the red half-edge at the bottom and if we replace it then the `GeometryGraph` will no longer be able to properly access the graph.

image

So when we split the edges from intersections we need to preserve the existing half-edges and change their `twin` and `next` values.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: Yes, but we need to at least have a blank entry in the `@geom_edges` array so that we can properly access each geometry in a collection.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: We should probably treat those the same way as points.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: There's no scenario I can think of where this would happen and the polygon would be valid, but I suppose that may not be the most robust test :)

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: After thinking about it more, I think it's fine to just leave it. It shouldn't affect anything else in the graph and just be disconnected from everything entirely. I'll add a test for it though.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: You're right I was able to reduce this to just `add_edges(edges)`

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: Nice! This works

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: Added a test, but this is handled smoothly.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: I changed it to a Set which I think is better and renamed the method to `cycle`. `cycle` is already an `Enumerable` method and I think it describes the operation better.

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: Sorry, I'm not sure exactly what you mean. Would you mind clarifying? I'll do this last since it will create a lot of diffs.

BuonOmo: I mean having the `PlanarGraph` class directly be parent of `HalfEdge`. Hence you would have:

- `RGeo::Cartesian::PlanarGraph`
- `RGeo::Cartesian::PlanarGraph::HalfEdge`
- `RGeo::Cartesian::GeometryGraph`

lib/rgeo/cartesian/planar_graph.rb
BuonOmo: `linked_edges` maybe? As cycle is also standard lib it might still be a bit confusing.

Although, we could make this protected and be sure users would not call this method anyway

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,382 @@
...
+            # Can't handle points yet, so just add an empty entry for them
+            @geom_edges << GeomEdge.new
+          when Feature::LineString, Feature::LinearRing
+            add_line_string(geom)
+          when Feature::Polygon
+            add_polygon(geom)
+          when Feature::GeometryCollection
+            add_collection(geom)
+          else
+            raise ArgumentError, "Invalid Geometry Type: #{geom.class}"
```
BuonOmo, line 292:
```suggestion
            raise Error::RGeoError, "Invalid Geometry Type: #{geom.class}"
```

We should only raise error that are children of `RGeoError`, to make sure someone can catch all of those if they want

lib/rgeo/cartesian/planar_graph.rb
BuonOmo: then can't we get read of the TODO? :)

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,382 @@
...
+          # segment of the ring.
+          start_seg = Segment.new(ring.start_point, ring.point_n(1))
+          end_seg = Segment.new(ring.point_n(ring.num_points - 2), ring.end_point)
+          colinear_hedges = hedges.select do |he|
+            start_seg.side(he.destination).zero? || end_seg.side(he.destination).zero?
+          end
+
+          colinear_hedges.find do |hedge|
+            pts = []
+            hedge.cycle do |he|
```
BuonOmo, line 371:
```suggestion
            hedge.and_connected do |he|
```

one more idea ^^

test/cartesian_planar_graph_test.rb
```diff
@@ -0,0 +1,442 @@
+# frozen_string_literal: true
+
+# -----------------------------------------------------------------------------
+#
+# Tests for the planar graph and related classes
+#
+# -----------------------------------------------------------------------------
+
+require "test_helper"
```
BuonOmo, line 9:
```suggestion
require_relative "test_helper"
```

This makes it so that we can `ruby thefile.rb` without need of any flag

BuonOmo (approved):
4 last comments, plus maybe the class renaming. I think it's good to me now, I'll let you finish :)

You can always slack me if something is unclear!

lib/rgeo/cartesian/planar_graph.rb
```diff
@@ -0,0 +1,378 @@
...
+          hedges << self
+
+          n = self.next
+          until hedges.include?(n) || n.nil?
+            yield(n) if block_given?
+            hedges << n
+            n = n.next
+          end
+          hedges
+        end
```
BuonOmo, line 77:
```suggestion
        # If a block is given, each HalfEdge seen will be yielded to the block, otherwise
        # returns an enumerator
        def and_connected
          return to_enum(__method__) unles block_given?

          hedges = Set.new
          yield(self)
          hedges << self

          n = self.next
          until hedges.include?(n) || n.nil?
            yield(n)
            hedges << n
            n = n.next
          end
        end
```

I guess you could always return an enum!

BuonOmo (approved):
LGTM! I've added one last review, nothing mandatory though :)

lib/rgeo/cartesian/planar_graph.rb
keithdoggett: Good idea! Done.

keithdoggett:
Thank you for all the reviews and help with this @BuonOmo. I'm going to merge this and get the validation checking PR up soon.

PR #270, opened 2021-07-31
Use native `GEOSEqualsExact_r` comparison

BuonOmo:
### Summary

This make for less C code on our part, and allowed to fix a MacOS
related bug.

### Other Information

At first I went with the `rgeo_geos_geometries_norm_eql` implementation, which allows to test for equality even if geometries are not in correct order. However, according to our documentation (https://github.com/rgeo/rgeo/blob/master/doc/An-Introduction-to-Spatial-Programming-With-RGeo.md#32-relational-operations), it seems that we should really use the `rgeo_geos_geometries_strict_eql` version, and just change the tests to use correct comparison. IMHO, it is ok to have possibly different representation when doing an unary union.

I won't keep both implementation when merging, I've just let those for review reasons!

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,6 +22,8 @@ jobs:
           - ubuntu
           - windows
           - macos
+        exclude:
+          - os: windows # See issue #242, windows support would be welcome.
     runs-on: ${{ matrix.os }}-latest
     continue-on-error: ${{ matrix.ruby == 'head' || matrix.os == 'windows' || matrix.os == 'macos' }}
     name: Ruby ${{ matrix.ruby }} (${{ matrix.os }})
--- a/History.md
+++ b/History.md
@@ -2,6 +2,7 @@
 
 * Change ProjectedLinearRing #is_simple? method to be uniform across geos versions #228
 * Improve large MultiPolygon creation performance (Quiwin) #251
+* Fix CAPI equality (`==`) comparison on macos by using `GEOSEqualsExact_r` #270
 
 ### 2.3.0 / 2021-04-16
 
--- a/ext/geos_c_impl/factory.h
+++ b/ext/geos_c_impl/factory.h
@@ -198,15 +198,6 @@ void rgeo_check_geos_object(VALUE obj);
 const GEOSGeometry* rgeo_get_geos_geometry_safe(VALUE obj);
 
 /*
-  Compares the coordinate sequences for two given GEOS geometries.
-  The two given geometries MUST be of types backed directly by
-  coordinate sequences-- i.e. points or line strings.
-  Returns Qtrue if the two coordinate sequences are equal, Qfalse
-  if they are inequal, or Qnil if an error occurs.
-*/
-VALUE rgeo_geos_coordseqs_eql(GEOSContextHandle_t context, const GEOSGeometry* geom1, const GEOSGeometry* geom2, char check_z);
-
-/*
   Compares the ruby classes and geometry factories of the two given ruby
   objects. Returns Qtrue if everything is equal (that is, the two objects
   are of the same type and factory), or Qfalse otherwise.
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -11,6 +11,7 @@
 #include <ruby.h>
 #include <geos_c.h>
 
+#include "errors.h"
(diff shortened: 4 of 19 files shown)


keithdoggett:
This looks good, definitely a lot simpler now. Does the `strict_eql` method require that points are sorted in the same order? For example would `POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))` and `POLYGON ((0 1, 1 1, 1 0, 0 0, 0 1))` not be considered equal?

I think it makes sense to check order of geometries when comparing collections, but if it comes at the expense of having the two above geometries considered not equal I'd probably prefer the normalized equality check.


FYI current equality check:

```rb
require 'rgeo'

fac = RGeo::Geos.factory
p1 = fac.parse_wkt("POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))")
p2 = fac.parse_wkt("POLYGON ((0 1, 1 1, 1 0, 0 0, 0 1))")

p p1 == p2
# => true
```

BuonOmo:
@keithdoggett just tested on my branch, and I get true as well for your given example. I've added that to the next tests:

```
rake TESTOPTS="--name='/(test_geometrically_equal_but_ordered_different|test_geometrically_equal_but_different_directions)/'"
```

keithdoggett:
Hmm, I ran a test in a C script and it looks like the ordering of points is affecting the equality.

```c
    // ...
    GEOSGeometry *gc5 = GEOSGeomFromWKT("POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))");
    GEOSGeometry *gc6 = GEOSGeomFromWKT("POLYGON ((0 1, 1 1, 1 0, 0 0, 0 1))");
    GEOSGeometry *gc7 = GEOSGeomFromWKT("POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))");
    GEOSGeometry *gc8 = GEOSGeomFromWKT("POLYGON ((0 1, 0 0, 1 0, 1 1, 0 1))");

    printf("%d\n", GEOSEqualsExact(gc5, gc6, 0.0));
    printf("%d\n", GEOSEqualsExact(gc5, gc7, 0.0));
    printf("%d\n", GEOSEqualsExact(gc5, gc8, 0.0));
    printf("%d\n", GEOSEqualsExact(gc6, gc7, 0.0));
    printf("%d\n", GEOSEqualsExact(gc5, gc5, 0.0));
```

Which prints
```
0
0
0
0
1
```

Do we normalize the points when a geometry is created? It seems like those tests you added shouldn't be passing.

BuonOmo:
The test is not failing because you are using strict equality, which corresponds to our `eql?` or `rep_equals?`, however it works with `==`:

```ruby
def polygon(pts, factory: RGeo::Geos.factory)
  factory.polygon(
    factory.linear_ring(
    )
  )
end

gc5 = polygon([0, 0,    0, 1,    1, 1,    1, 0,    0, 0])
gc6 = polygon([0, 1,    1, 1,    1, 0,    0, 0,    0, 1])
gc7 = polygon([0, 0,    1, 0,    1, 1,    0, 1,    0, 0])
gc8 = polygon([0, 1,    0, 0,    1, 0,    1, 1,    0, 1])

%i(== eql? rep_equals? equals? equal?).each do |cmp|
  puts
  puts "# Using #{cmp}"
  p [
    gc5.send(cmp, gc6),
    gc5.send(cmp, gc7),
    gc5.send(cmp, gc8),
    gc6.send(cmp, gc7),
    gc5.send(cmp, gc5)
  ]
end
```

```
# Using ==
[true, true, true, true, true]

# Using eql?
[false, false, false, false, true]

# Using rep_equals?
[false, false, false, false, true]

# Using equals?
[true, true, true, true, true]

# Using equal?
[false, false, false, false, true]
```

---

Here's the simplified call stack from ruby to GEOS:

| who | method |
| --- | ------ |
| rgeo (ruby) | == |
| rgeo (capi) | method_geometry_equals |
| geos | GEOSEquals_r |

And the `GEOSEquals_r` method is:

> * True if geometries cover the same space on the place.

keithdoggett:
Ahh gotcha that makes sense. I was just getting confused about where each was used. In that case, I think `strict_eql` should be used over `norm_eql` and once the `norm_eql` code is removed this should be good to merge. Thanks!

keithdoggett (approved):

PR #271, opened 2021-08-03
Add ImplHelper::ValidOp Module to Provide Uniform Validity Checks

keithdoggett:
### Summary

Implements `ImplHelper::ValidOp` which provides implementations of the `valid?` and `invalid_reason` methods based on simple geometric predicates.

Change summary:
- Implement `ImplHelper::ValidOp`
- Override `ValidOp` for cartesian polygons and use the `GeometryGraph` to implement the checks.
- Add `test/common/validity_test` to ensure uniform validity across all implementation types.
- Replace `validate_geometry` with `prepare_geometry` and move any relevant validations to the constructor.
  - These relevant validations are things like a `LinearRing` having enough points that GEOS will raise for so we have to also.
  - `prepare_geometry` will do the handy things that `validate_geometry` did like closing rings and ensuring points are in a projection area.

### Other Information

**Tests are currently failing because of the spherical factory**

There are still some open issues that need to be figured out before we can merge. The most pressing of which is the spherical implementation.

TODOS:

- [x] Implement geometric predicates in the spherical factory so that polygons can be validated (currently they cannot be).
- [ ] Make sure we like the structure of how classes can override their own checks from `ValidOp`. It seems a little clunky in cartesian polygons right now because a lot of the code is copied.
- [ ] Replace more of the overridden cartesian `ValidOp` methods with graph implementations. Once we've determined that the overriding structure makes sense I'll add more.
- [ ] Determine what to do with factory options like `uses_lenient_multipolygon_assertions` and `uses_lenient_assertions` (this may be part of @BuonOmo's validity PR that provides safe and unsafe versions of methods).
- [ ] Determine if all cartesian geometries should implement `graph` and re-implement some predicates with it (could be done later).


lib/rgeo/impl_helper/valid_op.rb
```diff
@@ -0,0 +1,357 @@
...
+
+        nil
+      end
+
+      ##
+      # Helper functions for specific validity checks
+      ##
+
+      def check_invalid_coordinate(pt)
+        return unless pt.x.nan? || pt.x.infinite? || pt.y.nan? || pt.y.infinite?
```
BuonOmo, line 196:
```suggestion
        return if pt.x.finite? && pt.y.finite?
```

Should we also check for rational (not sure if it is an issue) and complex (`Numeric#real?`) as well ?

lib/rgeo/cartesian/feature_methods.rb
```diff
@@ -50,31 +50,161 @@ def segments
...
-          oindex = nindex + 1
-          while oindex < len
-            oseg = segments[oindex]
-            return false if !(index == 0 && oindex == len - 1 && seg.s == oseg.e) && seg.intersects_segment?(oseg)
-            oindex += 1
-          end
-        end
-        true
+        # TODO: should we replace with graph.incident_edges.length == segments.length
+        # since graph isn't used elsewhere yet it might be more overhead.
```
BuonOmo, line 54: Not 100% sure I get your point there ^^

lib/rgeo/cartesian/feature_methods.rb
```diff
@@ -50,31 +50,161 @@ def segments
...
+        end
+
+        false
+      end
+    end
+
+    module PolygonMethods
+      def graph
+        @graph ||= GeometryGraph.new(self)
+      end
```
BuonOmo, line 104: Shouldn't this method be private api?

lib/rgeo/cartesian/feature_methods.rb
```diff
@@ -50,31 +50,161 @@ def segments
...
+
+      # Similar implementation to the default version from the ValidOp
+      # module, but it overrides a few methods in favor of a graph for
+      # easier checks. Eventually most of these checks can be overriden
+      # and rely on the graph.
+      #
+      # TODO: will update these with more checks once we decide
+      # if this structure makes sense or if there's more metaprogramming
+      # we could do.
+      def check_valid_polygon
```
BuonOmo, line 116: I don't understand why you copy this method: shouldn't the _leaf_ methods only be overriden?

I mean that since `check_consistent_area` and `check_connected_interiors` are overriden, you don't need to override that bit.

lib/rgeo/cartesian/feature_methods.rb
```diff
@@ -50,31 +50,161 @@ def segments
...
+      # exterior shell in the polygon's geometry graph. It will keep track
+      # of all the nodes it visited and if that set is a superset of the
+      # coordinates in the exterior_ring, the interior is connected, otherwise
+      # it is split somewhere.
+      #
+      # @return [String] invalid_reason
+      def check_connected_interiors
+        exterior_coords = exterior_ring.coordinates.to_set
+
+        visited = []
```
BuonOmo, line 198:
```suggestion
        visited = Set.new
```

just avoids an array object for nothing :)

lib/rgeo/cartesian/feature_methods.rb
```diff
@@ -50,31 +50,161 @@ def segments
...
-            oseg = segments[oindex]
-            return false if !(index == 0 && oindex == len - 1 && seg.s == oseg.e) && seg.intersects_segment?(oseg)
-            oindex += 1
-          end
-        end
-        true
+        # TODO: should we replace with graph.incident_edges.length == segments.length
+        # since graph isn't used elsewhere yet it might be more overhead.
+        li = SweeplineIntersector.new(segments)
+        li.proper_intersections.size.zero?
```
BuonOmo, line 56: eventually:
```suggestion
        li.proper_intersections.empty?
```

lib/rgeo/geos/ffi_feature_methods.rb
```diff
@@ -125,6 +125,15 @@ def valid?
         @fg_geom.valid?
       end
 
+      def invalid_reason
+        str = @fg_geom.valid_reason
+        if str.include?("Valid Geometry")
```
BuonOmo, line 130:
```suggestion
        if str == "Valid Geometry"
```

lib/rgeo/impl_helper/basic_line_string_methods.rb
```diff
@@ -176,16 +174,13 @@ def ccw?
         RGeo::Cartesian::Analysis.ccw?(self)
       end
 
-      private
+      # private
```
BuonOmo, line 177: I think this is a miss 🙂 (we should keep it, right?)

lib/rgeo/impl_helper/valid_op.rb
```diff
@@ -0,0 +1,357 @@
...
+      end
+      alias is_valid? valid?
+
+      # Reason for invalidity or nil if valid
+      #
+      # @return String
+      def invalid_reason
+        return @invalid_reason if defined?(@invalid_reason)
+        @invalid_reason = check_valid
+      end
```
BuonOmo, line 42: In my branch I do not memoize the reason, but rather have an `invalid_reason_memo` that does it. Anyway, I think we can leave that there for now, and discuss it in my PR when we'll have the complete flow!

test/common/validity_tests.rb
```diff
@@ -0,0 +1,217 @@
+# frozen_string_literal: true
+
+module RGeo
+  module Tests
+    module Common
+      module ValidityTests
+        def test_invalid_point_invalid_coordinate
+          assert(@invalid_point.invalid_reason.include?(RGeo::ImplHelper::TopologyErrors::INVALID_COORDINATE))
```
BuonOmo, line 8: I'm a bit unsure why you do not instantiate `@invalid_point` in the module? same for most instance variables there!

To do so, you could use memoization for instance

```ruby
def invalid_point
  return @invalid_point if defined?(@invalid_point)

  @invalid_point = ...
end

**Review by BuonOmo — COMMENTED** [review]
I think the way `ValidOp` is included is fine. As discussed in one of my comments, we can just override _leaf_ methods (i.e. the one that does the final check) to have something more accurate per implementation when needed.

If some method is not a _leaf_ and we need to override a part of it, we still can change the parent method, to make sure we always have a minimal overriding part.

Also, that has been a lot of work you've done on that validity stuff, thank you very much! It is a pleasure reading all of it 🙂

File: lib/rgeo/cartesian/feature_methods.rb (hunk 2 above)
Comment (keithdoggett) [reply lib/rgeo/cartesian/feature_methods.rb:54]: The same simplicity check (self-intersection) can be performed by creating the `GeometryGraph` of the `LineString` and doing the described comparison. But doing the check with `graph` will be slower since it has to perform an intersection check and then link all of the nodes whereas the current implementation only does an intersection check.

And since `graph` is not used elsewhere in any `LineString` method I figured the additional overhead is not necessary.

File: lib/rgeo/cartesian/feature_methods.rb (hunk 3 above)
Comment (keithdoggett) [reply lib/rgeo/cartesian/feature_methods.rb:104]: Yes, I also moved it to the `GeometryMethods` module so all cartesian geometries can create their graph.

File: lib/rgeo/cartesian/feature_methods.rb (hunk 4 above)
Comment (keithdoggett) [reply lib/rgeo/cartesian/feature_methods.rb:116]: The reason I did that is that in the originally implemented method in `ImplHelper` the leaf methods accept the geometry as an input, while these can just use self. But that was causing an issue where I'd be passing an argument into the new leaf methods that don't expect one.

I can change the leaf methods to something like this `def check_connected_interiors(_)` so that they won't break upon receiving an argument. What are your thoughts @BuonOmo?

File: lib/rgeo/impl_helper/valid_op.rb (hunk 1 above)
Comment (keithdoggett) [reply lib/rgeo/impl_helper/valid_op.rb]: I added the check for `real?` as well.

File: test/common/validity_tests.rb (hunk 10 above)
Comment (keithdoggett) [reply test/common/validity_tests.rb]: Good point. Done!

File: lib/rgeo/cartesian/feature_methods.rb [hunk 11]
```diff
@@ -50,31 +50,161 @@ def segments
...
+        interior_rings.each do |ring|
+          ring.points.each do |pt|
+            check = check_invalid_coordinate(pt)
+            return check unless check.nil?
+          end
+        end
+
+        # check closed
+        return ImplHelper::TopologyErrors::UNCLOSED_RING unless exterior_ring.is_closed?
+        return ImplHelper::TopologyErrors::UNCLOSED_RING unless interior_rings.all?(&:is_closed?)
```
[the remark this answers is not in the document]
keithdoggett: Yeah once it gets merged I can go back and change it.

lib/rgeo/cartesian/feature_methods.rb
keithdoggett: I think I'm doing something weird with the OO aspect of this. Since I'm including the `ValidOp` module every geometry instance will get these methods, but the leaf methods are designed to be more of a function than a method in the sense that they need the object they're checking passed in as a parameter and aren't checking the caller.

For example, a polygon calling `check_connected_interior` is essentially doing this

```rb
polygon.check_connected_interior(polygon)
```

where I feel like it should just be

```rb
polygon.check_connected_interior
```

Maybe I need to break this down into more submodules like `ValidOp::PolygonChecks` and include those into the relevant classes?

keithdoggett: I broke it up into `ValidOp` and `ValidOpHelpers` where the helpers contain the functions and `ValidOp` has just the `valid?`, `invalid_reason`, and `check_valid/check_valid_*` methods.

I also show how the helpers can be overridden for custom implementations of the leaf functions (`lib/cartesian/valid_op.rb`). The new helpers module has to "inherit" the default implementation and re-implement any methods it needs then the new valid op module defines a method which specifies the helper module it will make calls to. It is a bit of a weird pattern and I can clarify more if need be, but it seems to provide flexibility with minimal code duplication.

lib/rgeo/geographic/spherical_feature_methods.rb
```diff
@@ -123,6 +123,66 @@ def is_simple?
...
+        when Feature::LineString
+          crosses_line_string?(rhs)
+        else
+          super
+        end
+      end
+
+      private
+
+      # TODO: replace with better algorithm
```
BuonOmo, line 147: This should be translated into an issue I think :)

BuonOmo:
> Determine what to do with factory options like uses_lenient_multipolygon_assertions and uses_lenient_assertions (this may be part of @BuonOmo's validity PR that provides safe and unsafe versions of methods).

I'll definitely do that bit.

In fact I think we're ready to merge, since what is left is either optimisation, or spherical support. By merging we'll be able to split that job in many and have me and contributors help you :)

BuonOmo (approved):

lib/rgeo/geographic/spherical_feature_methods.rb
keithdoggett: #274

Issue #274, opened 2021-10-05
Spherical Math/Geometry Improvements


keithdoggett:
Currently, our implementation of spherical geometries lack lots of geometric predicates/analysis methods, are inefficient (ex. `SphericalLineString#crosses?` is O(n^2)), and are unstable when geometries have dense features (#212).

 TODOS:

- [ ] Change the `SphericalMath` module to improve stability for feature-dense geometries
- [ ] Add more methods to `SphericalMath` so that we can implement geometric methods more easily
- [ ] Implement more geometric methods on spherical geometries (`contains?`, `intersects?`, `union`, etc.).

This could be a helpful resource to borrow some ideas from https://spacetelescope.github.io/spherical_geometry/.

PR #275, opened 2021-10-05
Validity Handling

keithdoggett:
### Summary
Combination of #262 #269 and #271


keithdoggett:
GEOS 3.10 reworked their validity handling somewhat and the `DUPLICATE_RING` error is now folded in with `SELF_INTERSECTION`.

ext/geos_c_impl/geometry.c
```diff
@@ -1042,18 +1042,38 @@ static VALUE method_geometry_invalid_reason(VALUE self)
...
   if (self_geom) {
-    str = GEOSisValidReason_r(self_data->geos_context, self_geom);
-    // Per documentation, a valid geometry should give an empty string.
-    // However it seems not to be the case. Hence the comparison against
-    // the string that is really given: `"Valid Geometry"`.
-    // See https://github.com/libgeos/geos/issues/431.
-    if (str) result = (str[0] == '\0' || !strcmp(str, "Valid Geometry")) ? Qnil : rb_str_new2(str);
-    else result = rb_str_new2("Exception");
-    GEOSFree_r(self_data->geos_context, str);
+    // TODO: should we consider using the flag GEOSVALID_ALLOW_SELFTOUCHING_RING_FORMING_HOLE?
```
BuonOmo, line 1045: @keithdoggett since you implemented validity on the ruby part, may I have your opinion on that one?

lib/rgeo/cartesian/feature_methods.rb
```diff
@@ -50,26 +56,10 @@ def segments
...
-          next unless nindex
-          oindex = nindex + 1
-          while oindex < len
-            oseg = segments[oindex]
-            return false if !(index == 0 && oindex == len - 1 && seg.s == oseg.e) && seg.intersects_segment?(oseg)
-            oindex += 1
-          end
-        end
-        true
+        # TODO: should we replace with graph.incident_edges.length == segments.length
```
BuonOmo, line 59: @keithdoggett do you want to discuss that one?

lib/rgeo/geographic/spherical_feature_methods.rb
```diff
@@ -128,6 +128,66 @@ def is_simple?
...
+        when Feature::LineString
+          crosses_line_string?(rhs)
+        else
+          super
+        end
+      end
+
+      private
+
+      # TODO: replace with better algorithm
```
BuonOmo, line 152: I'm in favor of keeping it as-is, opening an issue if one wants to improve

lib/rgeo/impl_helper/valid_op.rb
```diff
@@ -0,0 +1,373 @@
...
+      INVALID_COORDINATE = "Invalid Coordinate"
+      UNCLOSED_RING = "Ring is not closed"
+    end
+
+    # Mixin based off of the JTS/GEOS IsValidOp class.
+    # Implements #valid? and #invalid_reason on Features that include this.
+    #
+    # @see https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/operation/valid/IsValidOp.java
+    module ValidOp
+      # TODO: determine if "Unkown Validity" should count as valid.
```
BuonOmo, line 29: I'd say it shouldn't, do you have examples @keithdoggett ?

keithdoggett: I agree that it shouldn't. If an error is raised along the way there is probably some sort of issue with the geometry anyways.

lib/rgeo/geographic/spherical_feature_methods.rb
keithdoggett: Do you want me to remove the TODOs or just leave them and open an issue?

lib/rgeo/cartesian/feature_methods.rb
keithdoggett: Yeah so when the `GeometryGraph` is instantiated, it uses a `SweeplineIntersector` to figure out intersections and help with noding of the graph. So the idea is that calling `graph.incident_edges.length == segments.length` is equivalent to running the `SweeplineIntersector` and checking if it's empty.

That being said, creating the `GeometryGraph` adds overhead since, in addition to finding the intersections, it also creates the graph of `HalfEdges`. `graph` is memoized, so if it's used multiple times it makes sense IMO to use graph here, but if `graph` is only used to determine simplicity, I think the reduced overhead of `SweeplineIntersector` makes sense.

Let me know if that's clear.

lib/rgeo/geographic/spherical_feature_methods.rb
BuonOmo: Either remove those or reference the issue in the TODOs, as you wish!

ext/geos_c_impl/geometry.c
BuonOmo: Well since users never had this option and no one requested it I'm not sure I'm in favor of that. The question was more : we have to pick one option (either always or never allow), which is the ruby counterpart? I'll try and find an example in the GEOS test suite to answer that question myself

BuonOmo:
```ruby
def ring(pts, factory: RGeo::Geos.factory)
  factory.linear_ring(
  )
end
[RGeo::Geos.factory, RGeo::Cartesian.simple_factory].each do |f|
	puts ring([0,0,	5,5,	0,10,	10,10,	5,5,	10,0,	0,0], factory: f).invalid_reason
end
```

Hence I'd be in favor of not adding the flag and sticking to SFS.

Another thing I found however that we may want to fix before releasing v3 is that when calling `make_valid` on the above ring we get `nil`. As we want to harmonize and always raise on error, I'd we should raise in that particular case. What do you say?

keithdoggett: Ahh, gotcha yes, I agree that we should not use that flag and the ring self-intersections be treated as invalid since that is not standard.

Good catch with `make_valid` return `nil`. Should definitely raise if it can't make it valid.

lib/rgeo/geographic/spherical_feature_methods.rb
keithdoggett: Already created #274 which covers this. I'll add that to the TODO

lib/rgeo/cartesian/feature_methods.rb
keithdoggett: Left a note about it.

keithdoggett:
@BuonOmo thanks for writing that doc. I added a section at the end that clarifies which types of methods require a validity check first.

BuonOmo:
Merging into v3-dev now since the workflow is satisfying, issues left will be adressed per PR

Issue #277, opened 2021-11-03
 RGeo::Geographic.simple_mercator_factory ignores lenient flags


earnold:
I am having an issue with the classic `LinearRing failed ring test` when I use a simple mercator factory.  I've noticed that several of the closed issues on this topic suggest setting one of the possible lenient flags, i.e. `lenient_multi_polygon_assertions: true, uses_lenient_assertions:true`

However, this did not solve my issue. Looking under the hood, it seems as if [the factor method does not pass along these options](https://github.com/rgeo/rgeo/blob/v2.3.0/lib/rgeo/geographic/interface.rb#L216). 

Is this a deliberate design choice? And if so, could anyone point me in the direction of getting a more lenient behavior with a simple mercator projection & a potentially self-intersecting multipolygon?

earnold:
I've posted more details about a related issue on stackoverflow if that is helpful: https://stackoverflow.com/questions/69827860/issue-with-rgeo-geojson-linearring-failed-ring-test-epsg-3857-vs-epsg-432

thestelz:
@earnold I've had the same issue and I have created the following monkey patch. I was going to put in a pull request about fixing this, but it they are in the progress of revamping the gem, so I wanted to wait.

I pulled this from my project, so it should work as a drop in replacement. 

```ruby
# Monkey patch for issues with lenient assertions for RGeo.
#
# /lib/extensions/rgeo/factory.rb
#
# ruby '2.5.3'
# gem 'rails', '5.2.4.6'
# gem 'rgeo', '2.1.1'
#
# This will allow:
#   factory = RGeo::Geographic.simple_mercator_factory(uses_lenient_assertions: true, lenient_multi_polygon_assertions: true)
#   factory.set_property(:uses_lenient_assertions, true)
#
module RGeo
  module Geographic
    class Factory
      def set_property(prop, value)
        case prop
          when :has_z_coordinate
            @support_z = value
          when :has_m_coordinate
            @support_m = value
          when :uses_lenient_assertions
            @lenient_assertions = value
          when :buffer_resolution
            @buffer_resolution = value
          when :is_geographic
            value
        end
      end
    end
  end
end
```

I hope this helps!

(I'll also post on the S.O. question as well)

earnold:
@thestelz Awesome! Thank you. This is helpful.

keithdoggett:
[...]

PR #278, opened 2021-11-05
fix: lenient_assertion flags for simple_mercator_factory

thestelz:
Fix `:uses_lenient_assertions` flag for
`RGeo::Geographic.simple_mercator_factory`.

fixes rgeo/rgeo#277

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ mkmf.log
 Gemfile.lock
 /yardoc
 /.yardoc
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ mkmf.log
 Gemfile.lock
 /yardoc
 /.yardoc
+.idea
--- a/lib/rgeo/geographic/interface.rb
+++ b/lib/rgeo/geographic/interface.rb
@@ -222,7 +222,8 @@ module RGeo
           wkt_generator: opts[:wkt_generator],
           wkb_generator: opts[:wkb_generator],
           has_z_coordinate: opts[:has_z_coordinate],
-          has_m_coordinate: opts[:has_m_coordinate])
+          has_m_coordinate: opts[:has_m_coordinate],
+          uses_lenient_assertions: opts[:uses_lenient_assertions])
         projector = Geographic::SimpleMercatorProjector.new(factory,
           buffer_resolution: opts[:buffer_resolution],
           lenient_multi_polygon_assertions: opts[:lenient_multi_polygon_assertions],
--- a/test/simple_mercator/factory_test.rb
+++ b/test/simple_mercator/factory_test.rb
@@ -15,4 +15,10 @@ class MercatorFactoryTest < Minitest::Test # :nodoc:
     @factory = RGeo::Geographic.simple_mercator_factory
     @srid = 4326
   end
+
+  def test_has_uses_lenient_assertions
+    factory = RGeo::Geographic.simple_mercator_factory(uses_lenient_assertions: true)
+
+    assert_equal(true, factory.property(:uses_lenient_assertions))
+  end
 end
--- a/test/simple_mercator/factory_test.rb
+++ b/test/simple_mercator/factory_test.rb
@@ -23,14 +23,14 @@ class MercatorFactoryTest < Minitest::Test # :nodoc:
   end
 
   def test_ring_has_lenient_assertions
-    ring = -> (factory) {
+    ring = lambda do |factory|
       factory.linear_ring([
-        factory.point(0, 0),
-        factory.point(1, 1),
-        factory.point(0, 1),
-        factory.point(1, 0)
-      ])
-    }
+                            factory.point(0, 0),
+                            factory.point(1, 1),
+                            factory.point(0, 1),
+                            factory.point(1, 0)
+                          ])
+    end
 
     assert_raises { ring.call RGeo::Geographic.simple_mercator_factory }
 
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -14,3 +14,4 @@ mkmf.log
 Gemfile.lock
 /yardoc
 /.yardoc
+.idea
```
keithdoggett, line 17: Can we get rid of this? I'd prefer to keep the IDE related stuff out of here.

keithdoggett (changes requested):
Thanks for the PR! Would you be able to add a test in `test/simple_mercator/factory_test.rb` that tests these changes work?

thestelz:
> Thanks for the PR! Would you be able to add a test in `test/simple_mercator/factory_test.rb` that tests these changes work?

I'm not sure how to fully test for this since there isn't access to that boolean flag. This is what I have so far, but not sure where to go from here, so any help/info/tips would be appreciated. 

```ruby
# frozen_string_literal: true

# -----------------------------------------------------------------------------
#
# Tests for the GEOS factory
#
# -----------------------------------------------------------------------------

require "test_helper"

class MercatorFactoryTest < Minitest::Test # :nodoc:
  include RGeo::Tests::Common::FactoryTests

  def setup
    @factory = RGeo::Geographic.simple_mercator_factory
    @srid = 4326
  end

  def test_has_uses_lenient_assertions
    factory = RGeo::Geographic.simple_mercator_factory(uses_lenient_assertions: true)

    # TODO: assert_equal(true, factory.uses_lenient_assertions)
  end
end

```

thestelz:
>You can use the `property` method on the factory to get it.

Thanks! Sorry, switching between three different projects and this one I get lost on what methods I can call sometimes.

keithdoggett:
No problem, there's quite a few methods and classes so it's pretty easy to get mixed up. Thanks for fixing that quickly

keithdoggett (approved):
LGTM! Letting @BuonOmo have a look then we can merge

.gitignore
BuonOmo: Definitely! @thestelz note that you can have a global ignore file for your own configuration:

```bash
echo '.idea' > ~/.gitignore 
git config --global core.excludesFile ~/.gitignore
```

See http://git-scm.com/docs/gitignore#_description for more

BuonOmo (approved):
I only wonder if we should add a test related to creating a linear ring without risks using lenient assertion. Something like:

```ruby
ring = -> (factory) {
  factory.linear_ring([
    factory.point(0, 1),
    factory.point(1, 1),
    factory.point(1, 0)
  ])
}
assert_raises { ring.call RGeo::Geographic.simple_mercator_factory }
ring.call RGeo::Geographic.simple_mercator_factory(uses_lenient_assertions: true)
```

Might not be necessary though, LGTM anyway. Thank you for your contribution!

.gitignore
thestelz: @BuonOmo sorry, we all use the same ide, so it is habit to add it before committing. We don't use the global due to new hires needed to know to do that, so one less thing to slip through the cracks. I will, however, add it to mine so this doesn't happen in the future. Thanks for the info!

thestelz:
I might be missing something, but I added the follow to the test:

```ruby
  def test_ring_has_lenient_assertions
    ring = -> (factory) {
      factory.linear_ring([
        factory.point(0, 1),
        factory.point(1, 1),
        factory.point(1, 0)
      ])
    }

    assert_raises { ring.call RGeo::Geographic.simple_mercator_factory }

    ring.call RGeo::Geographic.simple_mercator_factory(uses_lenient_assertions: true)
  end
```

and I'm getting:

```shell
Minitest::Assertion: StandardError expected but nothing was raised.
/rgeo/test/simple_mercator/factory_test.rb:34:in `test_ring_has_lenient_assertions'
```

Do I need to change your above code to get it to run correctly?

keithdoggett:
@thestelz sorry about the delay getting back to you.

I think the test should be this

```ruby
    def test_ring_has_lenient_assertions
      ring = -> (factory) {
        factory.linear_ring([
          factory.point(0, 0),
          factory.point(1, 1),
          factory.point(0, 1),
          factory.point(1, 0)
        ])
      }
  
      assert_raises { ring.call RGeo::Geographic.simple_mercator_factory }
  
      ring.call RGeo::Geographic.simple_mercator_factory(uses_lenient_assertions: true)
    end
```

since that is an invalid geom that we would expect raises an error.

thestelz:
@keithdoggett Thanks for that! It is working like a charm now. I have committed and pushed everything.

BuonOmo:
Sorry to bother you one last time @thestelz, rubocop is now failing :/

```
test/simple_mercator/factory_test.rb:26:12: C: [Correctable] Layout/SpaceInLambdaLiteral: Do not use spaces between -> and ( in lambda literals.
    ring = -> (factory) {
           ^^^^^^^^^^^^
test/simple_mercator/factory_test.rb:26:12: C: [Correctable] Style/Lambda: Use the lambda method for multiline lambdas.
    ring = -> (factory) {
           ^^
test/simple_mercator/factory_test.rb:28:9: C: [Correctable] Layout/FirstArrayElementIndentation: Use 2 spaces for indentation in an array, relative to the first position after the preceding left parenthesis.
        factory.point(0, 0),
        ^^^^^^^^^^^^^^^^^^^
test/simple_mercator/factory_test.rb:32:7: C: [Correctable] Layout/FirstArrayElementIndentation: Indent the right bracket the same as the first position after the preceding left parenthesis.
      ])
      ^
```

would you be so kind to fix that? I'll push a fix otherwise :)

thestelz:
@BuonOmo I have updated the styling issues. Sorry, forgot to run `rubocop` on my changes before the last update.

keithdoggett:
Thanks for the work on this! I'll get this released later today.

thestelz:
NP! Thanks for the help in getting it fixed.

Issue #273, opened 2021-08-30
Change CAPI data wrappers


BuonOmo:
`Data_Wrap_Struct` has been deprecated a long time ago, we want to replace it with the `Typed..` macros.

This refacto should only make the code simpler and will be part of a major with breaking changes, feel free to change the structure (and discuss the changes here beforehand if you are unsure).

See https://github.com/ruby/ruby/blob/HEAD/doc/extension.rdoc

PR #279, opened 2021-11-30
Update CAPI wrappers to ditch deprecated Data_Wrap_Struct

teckwan:
### Summary

Addresses https://github.com/rgeo/rgeo/issues/273, replacing `Data_Wrap_Struct` with its typed counterpart.

### Other Information

This is my first contribution in RGeo, and since I'm unfamiliar with the code styles, the codebase, and coding extension libraries in general, please don't hesitate to be critical and address even minor changes or cosmetic ones.

### DIFF
--- a/ext/geos_c_impl/factory.c
+++ b/ext/geos_c_impl/factory.c
@@ -431,7 +431,7 @@ static VALUE cmethod_factory_create(VALUE klass, VALUE flags, VALUE srid, VALUE 
       data->wkrep_wkb_parser = Qnil;
       data->proj4_obj = proj4_obj;
       data->coord_sys_obj = coord_sys_obj;
-      result = Data_Wrap_Struct(klass, mark_factory_func, destroy_factory_func, data);
+      result = TypedData_Wrap_Struct(klass, &rgeo_factory_type, data);
     }
     else {
       free(data);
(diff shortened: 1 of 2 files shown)


BuonOmo:
@keithdoggett FYI, Teck is a coworker at Klaxit, I've paired reviewed the code already, and there will be some slight cosmetic changes, apart from that LGTM. I'll let you review after that :)

BuonOmo (approved):
LGTM!

keithdoggett (approved):
@teckwan thanks so much for your help with this! LGTM I just have a few questions.

1) Should we add the ability to mark these as movable, like we did in the proj4 extension (https://github.com/rgeo/rgeo-proj4/blob/3b55631538b54fd64d7410dd0e2e10be0280235f/ext/proj4_c_impl/main.c#L103-L119)? I figure this would be a good time to do it.

2) Should this be merged into v3-dev instead? Do you have any thoughts around that @BuonOmo?

BuonOmo:
@keithdoggett 

On (1.) I definitely agree, yet this could be part of another PR since tests will be tricky.

On (2.) the only issue would be for apps building around private parts of our c api, I think this can be part of a patch without any issue.

keithdoggett:
Ok cool so this should be good to merge then and we can add support for marking in a different PR then.

keithdoggett:
Thanks @teckwan!

PR #282, opened 2022-01-07
Refactor globals and GC mark them

teckwan:
### Summary

Refactor the globals, to make them global in C, and remove the link between factory and globals. This avoids passing references around, and treating ruby as a sort of storage for these globals.

### Reproduction

The script below would segv every time (every rgeo / geos version that can compile nowadays)

```ruby
pt=RGeo::Geos.factory.point(1,0)
RGeo::Geos::CAPIFactory::INTERNAL_CGLOBALS = nil
def polygon(pts, factory: RGeo::Geos.factory)
  factory.polygon(
    ring(pts, factory: factory)
  )
end
polygon([0, 0,    0, 1,    1, 1,    1, 0,    0, 0]).contains?(pt)
```

Even though this is a pretty dumb script, pure ruby should not crash, moreover, using GC.compact a few times (or `verify_compaction_references`) would generate the same issue at some point.

### DIFF
--- a/ext/geos_c_impl/factory.c
+++ b/ext/geos_c_impl/factory.c
@@ -10,6 +10,8 @@
 #include <ruby.h>
 #include <geos_c.h>
 
+#include "globals.h"
+
 #include "factory.h"
 #include "geometry.h"
 #include "point.h"
@@ -130,23 +132,6 @@ static void mark_geometry_func(RGeo_GeometryData* data)
 }
 
 
-// Destroy function for globals data. We don't need to destroy any
-// auxiliary data for now...
-
-static void destroy_globals_func(RGeo_Globals* data)
-{
-  free(data);
-}
-
-
-// Mark function for globals data. This should mark any globals that
-// need to be held through garbage collection (none at the moment.)
-
-static void mark_globals_func(RGeo_Globals* data)
-{
-}
-
-
 /**** RUBY METHOD DEFINITIONS ****/
 
 
@@ -248,7 +233,6 @@ static VALUE method_factory_read_for_marshal(VALUE self, VALUE str)
   return result;
 }
 
-
 static VALUE method_factory_read_for_psych(VALUE self, VALUE str)
 {
   RGeo_FactoryData* self_data;
@@ -275,6 +259,9 @@ static VALUE method_factory_read_for_psych(VALUE self, VALUE str)
   return result;
 }
 
+#ifndef RGEO_GEOS_SUPPORTS_SETOUTPUTDIMENSION
+static VALUE marshal_wkb_generator;
+#endif
 
 static VALUE method_factory_write_for_marshal(VALUE self, VALUE obj)
 {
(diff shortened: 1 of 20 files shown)


ext/geos_c_impl/factory.c
```diff
@@ -929,7 +848,8 @@ VALUE rgeo_geos_klasses_and_factories_eql(VALUE obj1, VALUE obj2)
   }
   else {
     factory = RGEO_GEOMETRY_DATA_PTR(obj1)->factory;
-    result = rb_funcall(factory, RGEO_FACTORY_DATA_PTR(factory)->globals->id_eql, 1, RGEO_GEOMETRY_DATA_PTR(obj2)->factory);
+    /* No need to cache the internal here (https://ips.fastruby.io/4x) */
+    result = rb_funcall(factory, rb_intern("eql?"), 1, RGEO_GEOMETRY_DATA_PTR(obj2)->factory);
```
BuonOmo, line 852: We extended the usage of this benchmark to `"hash"` and `"cast"` as well, the idea being that their usage does not call for such precise performance optimisation (otherwise, we should not be calling `rb_funcall` either).

Performance matters, however we're working with geometries that are much computation heavier than featching a value in a global symbol cache, hence we decided to pick our fights and prefer readability here.

The benchmark was made with:

```ruby
require "benchmark/ips"
load "#{__dir__}/lib/rgeo.rb"

# Puts a Centered, bold and inverted text for better visibility.
def shout_out(string)
  puts "\n\e[7;1m#{string.center(75)}\e[27;0m\n\n"
end

shout_out("SMALL POLYGON")

def ring(pts, factory: RGeo::Geos.factory)
  factory.linear_ring(
  )
end

def polygon(pts, factory: RGeo::Geos.factory)
  factory.polygon(
    ring(pts, factory: factory)
  )
end

poly = polygon([0, 0,    0, 1,    1, 1,    1, 0,    0, 0])

Benchmark.ips do |x|
  x.config(time: 20)
  x.report("rb_intern") { poly == poly }
  x.report("cached") { poly == poly }
  x.hold!("temp")
  x.compare!
end
```

and commenting the polygon equality part to only have this class used in c.

ext/geos_c_impl/factory.c
```diff
@@ -634,33 +576,16 @@ RGeo_Globals* rgeo_init_geos_factory()
   rb_define_module_function(geos_factory_class, "_supports_unary_union?", cmethod_factory_supports_unary_union, 0);
 
   // Pre-define implementation classes and set up allocation methods
```
BuonOmo, line 578: @teckwan I think we should update this comment

ext/geos_c_impl/factory.c
```diff
@@ -339,25 +322,18 @@ static VALUE method_factory_write_for_psych(VALUE self, VALUE obj)
...
   has_3d = self_data->flags & RGEO_FACTORYFLAGS_SUPPORTS_Z_OR_M;
 #ifndef RGEO_GEOS_SUPPORTS_SETOUTPUTDIMENSION
   if (has_3d) {
-    globals = self_data->globals;
-    wkt_generator = globals->psych_wkt_generator;
-    if (NIL_P(wkt_generator)) {
-      wkt_generator = rb_funcall(
-        rb_const_get_at(globals->geos_module, rb_intern("Utils")),
+    if (NIL_P(psych_wkt_generator)) {
+      psych_wkt_generator = rb_funcall(
```
BuonOmo, line 332: shouldn't this be marked as well ? Note that however, in the current case, I did not find a way to generate a segv.

I tried some variations of :

```ruby
RGeo::Geos::Utils.singleton_class.undef_method :psych_wkt_generator
RGeo::WKRep::WKTGenerator.undef_method :generate
RGeo::Geos::Utils = nil
f = RGeo::Geos.factory(has_z_coordinate: true)
pt = f.point(1, 0, 0);nil
GC.start
GC.verify_compaction_references(double_heap: true, toward: :empty)
f.write_for_psych(pt)

ObjectSpace.each_object(RGeo::WKRep::WKTGenerator).count # run this between every line to see evolution
```

At some point, I just found that `RGeo::Geos.capi_supported?` is false. It looks like there is an issue with our new organisation:

image

teckwan: Effectively there was a need to mark this as well. However i think that since the value mutates, simply using `rb_gc_register_mark_object` was not sufficient (still segfaults using ur reproduction). By using `rb_gc_register_address` it seems to be better.

ext/geos_c_impl/factory.c
```diff
@@ -569,48 +542,19 @@ static VALUE alloc_geometry(VALUE klass)
...
-  globals->sym_force_new = ID2SYM(rb_intern("force_new"));
-  globals->sym_keep_subtype = ID2SYM(rb_intern("keep_subtype"));
+
 #ifndef RGEO_GEOS_SUPPORTS_SETOUTPUTDIMENSION
-  globals->psych_wkt_generator = Qnil;
-  globals->marshal_wkb_generator = Qnil;
+  psych_wkt_generator = Qnil;
+  rb_gc_register_address(&psych_wkt_generator);
+  marshal_wkb_generator = Qnil;
+  rb_gc_register_address(&marshal_wkb_generator);
```
BuonOmo, line 553:
```suggestion
  /* We favor rb_gc_register_address over rb_gc_register_mark_object because the value changes at runtime */
  psych_wkt_generator = Qnil;
  rb_gc_register_address(&psych_wkt_generator);
  marshal_wkb_generator = Qnil;
  rb_gc_register_address(&marshal_wkb_generator);
```

BuonOmo (approved):
One last comment to add IMHO (reflecting the last answer you gave me) otherwise, LGTM!

@keithdoggett could you review this one as well ?

keithdoggett (approved):
This looks great thanks for the work on this! The notes mentioned in there make sense and this seems a lot better than passing around the globals struct.

ext/geos_c_impl/factory.c
keithdoggett: yeah that makes sense to add a note. I didn't realize you could mark objects with the `rb_gc_register_address` and `rb_gc_mark` functions. Pretty cool

PR #283, opened 2022-01-13
V3

BuonOmo:
### Summary

- validity handling

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,7 @@
+### Current
+
+* Change `ProjectedLinearRing#is_simple?` method to be uniform across geos versions #228
+
 ### 2.3.1 / 2021-11-30
 
 * Fix core dump when cast returns a non Geos object #261
@@ -18,7 +22,6 @@
 * Add a simple fallback for `Polygon#contains?` (Quiwin) #224
 * Add `ccw?` method for linear rings, with geos support #229
 * Add a simple fallback for `MultiPolygon/LineString#contains?` (Quiwin) #230 #232
-
 ### 2.1.1 / 2019-8-26
 
 * Fix BasicPolygonMethods#boundary #206 (ans82)
--- a/History.md
+++ b/History.md
@@ -1,6 +1,7 @@
 ### Current
 
 * Change `ProjectedLinearRing#is_simple?` method to be uniform across geos versions #228
+* Improve large MultiPolygon creation performance (Quiwin) #251
 
 ### 2.3.1 / 2021-11-30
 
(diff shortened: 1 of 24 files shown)


doc/Geometry-Validity.md
```diff
@@ -0,0 +1,92 @@
+# Geometry Validity
+
+**TL;DR:**
+
+```ruby
+factory = RGeo::Cartesian.preferred_factory
+bowtie_polygon = factory.polygon(factory.linear_ring([factory.point(0, 4),factory.point(4, 4),factory.point(2, 6),factory.point(0, 0),factory.point(0, 4)]))
+polygon =
+	case bowtie_polygon.invalid_reason
+	when RGeo::ImplHelper::TopologyErrors::SELF_INTERSECTION
```
BuonOmo, line 10: @keithdoggett do we want to simplify the path for this kind of usage? It feels weird to tell a rgeo user to call something under `ImplHelper`

keithdoggett: Yeah that's smart. Do you think I should put it in the errors module so it's `RGeo::Errors::TopologyErrors` or just make it `RGeo::TopologyErrors` and put it in the top level of the `valid_op.rb` file?

BuonOmo: I'd go for the first one since every error is under `RGeo::Errors` already, can do it tomorrow

lib/rgeo/geographic/projected_feature_methods.rb
```diff
@@ -154,7 +167,7 @@ def self.included(klass)
 
       private
 
-      def validate_geometry
+      def prepare_geometry
```
keithdoggett, line 170: I wonder if I should change instances of `prepare_geometry` to another name. `prepare_geometry` is responsible for things like automatically closing linear rings or making sure that coordinates are on the globe, but it could be confused with functionality related to a prepared geometry (https://libgeos.org/usage/c_api/#prepared-geometry). Since this method is private and is only used in constructors of geometries it might not matter much though.

BuonOmo: @keithdoggett maybe `init_geometry` and a comment ? :)

History.md
```diff
@@ -1,4 +1,4 @@
-### Current
+### 3.0.0
```
BuonOmo, line 1:
```suggestion
### 3.0.0-rc.1
```

keithdoggett (approved):

Issue #280, opened 2021-12-07
Make CAPI wrapped pointers movable


BuonOmo:
> Should we add the ability to mark these as movable, like we did in the proj4 extension (https://github.com/rgeo/rgeo-proj4/blob/3b55631538b54fd64d7410dd0e2e10be0280235f/ext/proj4_c_impl/main.c#L103-L119)? I figure this would be a good time to do it.

_Originally posted by @keithdoggett in https://github.com/rgeo/rgeo/pull/279#pullrequestreview-822126976_

This should be tested using `GC.compact` or moving specific objects.

Useful resources:

- https://github.com/ohler55/oj/pull/653 another example
- https://www.youtube.com/watch?v=H8iWLoarTZc
- https://github.com/tenderlove/heap-utils

BuonOmo:
Fixed by #284

PR #284, opened 2022-01-14
Movable pointers

teckwan:
### Summary

Addresses https://github.com/rgeo/rgeo/issues/280, marking pointers as movable to allow for compaction.
Tested with the following scripts :

<details>
<summary> Scripts used </summary>

To dump `ObjectSpace`:
```ruby
# frozen_string_literal: true

require "objspace"
require_relative "./lib/rgeo"

ObjectSpace.trace_object_allocations_start

def record_compaction
  3.times { GC.start }
  GC.compact

  File.open("compacted.json", "w") do |file|
    ObjectSpace.dump_all(output: file)
  end
end

class Allocator
  def initialize
    @padding = []
    @held = []
  end

  def padding(n = 5_000)
    n.times do
      @padding << allocate_geometry
    end
  end

  def hold(n = 1_000)
    n.times do
      @held << allocate_geometry
    end
  end

  def allocate_factory
    RGeo::Geos::CAPIFactory.create(wkt_parser: {})
  end

  def allocate_geometry
    factory = allocate_factory

    ring = factory.linear_ring(
      [
        factory.point(rand(0..5), rand(0..5)),
        factory.point(rand(6..10), rand(0..5)),
        factory.point(rand(6..10), rand(6..10)),
        factory.point(rand(0..5), rand(6..10))
      ]
    )

    factory.polygon(ring)
  end

  def release_padding
    @padding.clear
  end
end

a = Allocator.new

a.padding
a.hold
a.release_padding

record_compaction
```

To visualize them : a modified version o f https://github.com/tenderlove/heap-utils
</details>

Results do seem positive, the graphics below correspond to the heap, with red dots as the pinned objects. We can see an overall decrease in pinned objects.

For geometries :
before patch | after patch
----|----
8543 pinned | 3538 pinned
before-patch-geometry | before

For factories :
before patch | after patch
----|----
7545 pinned | 3537 pinned
before-patch | after-patch

As for having tests to ensure non-regression, `GC.auto_compact` is enabled (option available only on ruby 3). Major collections will be done at least once during a test suite, which will trigger compaction. By commenting out the source code in `compact_*` functions, I am encountering a SEGV once every 10 test runs roughly. This is far from being an ideal test case but can serve as a simple way to be eventually aware of a regression.

### Other Information

The last commit is to remove warnings on compilation I have precedently introduced. Do tell me if you want it elsewhere.

### DIFF
--- a/ext/geos_c_impl/factory.c
+++ b/ext/geos_c_impl/factory.c
@@ -38,82 +38,89 @@ static void message_handler(const char* fmt, ...)
 // objects that have been created for the factory, and then destroy
 // the GEOS context, before freeing the factory data itself.
 
-static void destroy_factory_func(RGeo_FactoryData* data)
+static void destroy_factory_func(void* data)
 {
+  RGeo_FactoryData* factory_data;
   GEOSContextHandle_t context;
 
-  context = data->geos_context;
-  if (data->wkt_reader) {
-    GEOSWKTReader_destroy_r(context, data->wkt_reader);
+  factory_data = (RGeo_FactoryData*)data;
+  context = factory_data->geos_context;
+  if (factory_data->wkt_reader) {
+    GEOSWKTReader_destroy_r(context, factory_data->wkt_reader);
   }
-  if (data->wkb_reader) {
-    GEOSWKBReader_destroy_r(context, data->wkb_reader);
+  if (factory_data->wkb_reader) {
+    GEOSWKBReader_destroy_r(context, factory_data->wkb_reader);
   }
-  if (data->wkt_writer) {
-    GEOSWKTWriter_destroy_r(context, data->wkt_writer);
+  if (factory_data->wkt_writer) {
+    GEOSWKTWriter_destroy_r(context, factory_data->wkt_writer);
   }
-  if (data->wkb_writer) {
-    GEOSWKBWriter_destroy_r(context, data->wkb_writer);
+  if (factory_data->wkb_writer) {
+    GEOSWKBWriter_destroy_r(context, factory_data->wkb_writer);
   }
-  if (data->psych_wkt_reader) {
-    GEOSWKTReader_destroy_r(context, data->psych_wkt_reader);
+  if (factory_data->psych_wkt_reader) {
+    GEOSWKTReader_destroy_r(context, factory_data->psych_wkt_reader);
   }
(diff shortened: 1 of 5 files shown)


BuonOmo:
This looks really promising, as we've been work a lot in pair on this one I think @keithdoggett will be a better reviewer than me on that topic.

LGTM

Failing tests are related to geometry comparison, which we fixed in v3-dev

ext/geos_c_impl/factory.c
```diff
@@ -38,100 +38,175 @@ static void message_handler(const char* fmt, ...)
...
+  RGeo_GeometryData* geometry_data;
+
+  geometry_data = (RGeo_GeometryData*)data;
+  if (!NIL_P(geometry_data->factory)) {
+    geometry_data->factory = rb_gc_location(geometry_data->factory);
+  }
+  if (!NIL_P(geometry_data->klasses)) {
+    geometry_data->klasses = rb_gc_location(geometry_data->klasses);
+  }
+}
```
keithdoggett, line 183: Since I'm not extremely familiar with all of the memory management internals could you explain how the `compact_` method is different than the `mark_` and why we need both?

I'm guessing compact is run before an object is being moved, but just want to clarify since they seem very similar.

keithdoggett (approved):
LGTM! I left a little note with a question about memory management I have, but this looks great thanks @teckwan and @BuonOmo

ext/geos_c_impl/factory.c
teckwan: They're actually more or less two pieces of the same puzzle. We are using the movable variant `rb_gc_mark_movable` which tells the GC that this object can be moved. When the GC actually moves this object, it will then leave a trail in the form of a pointer, so the object does not get lost. As the result of this, we end up with pointers pointing other pointers. This is where `compact_` comes in, where we define the logic to remove the unnecessary pointers. @keithdoggett

keithdoggett: Gotcha that makes sense thanks for the explanation @teckwan

keithdoggett:
LGTM! Merge whenever you're ready @BuonOmo. Are you going to roll this into a release as well?

BuonOmo:
@keithdoggett @teckwan it is now part of RGeo 2.4.0 :rocket:

PR #286, opened 2022-01-31
Add error handler for geos

BuonOmo:
### Summary

When coding for the validity handling we noticed that there were still some places where we return `nil` rather than raising for invalidity or unsupported operations. This PR fixes it.

### TODO

Tests are not yet passing, there are some errors here that should be categorised correctly. On the notice handler part, when running tests with `DEBUG=1 rake` we can see that there are a lot of warnings, surely something to handle

### DIFF
--- a/doc/Factory-Compatibility.md
+++ b/doc/Factory-Compatibility.md
@@ -408,7 +408,7 @@ _Note: This list is not exhaustive of all the methods defined by each factory. T
 |                         `Collection#as_binary` |  ✅   |    ✅    |  ✅  |   ✅    |     ✅     |     ✅      |     ✅     |
 |                            `Collection#empty?` |  ✅   |    ✅    |  ✅  |   ✅    |     ✅     |     ✅      |     ✅     |
 |                           `Collection#simple?` |  ✅   |    ✅    |  ✅  |   ✅    |     ❌     |     ✅      |     ❌     |
-|                          `Collection#boundary` |  ✅   |    ❌    |  ✅  |   ❌    |     ❌     |     ✅      |     ❌     |
+|                          `Collection#boundary` |  ❌   |    ❌    |  ❌  |   ❌    |     ❌     |     ❌      |     ❌     |
 |                       `Collection#convex_hull` |  ✅   |    ✅    |  ✅  |   ✅    |     ❌     |     ✅      |     ❌     |
 |                            `Collection#buffer` |  ✅   |    ✅    |  ✅  |   ✅    |     ❌     |     ✅      |     ❌     |
 |                    `Collection#equals?(Point)` |  ✅   |    ✅    |  ✅  |   ✅    |     ❌     |     ✅      |     ❌     |
--- a/lib/rgeo/geos/ffi_feature_methods.rb
+++ b/lib/rgeo/geos/ffi_feature_methods.rb
@@ -147,6 +147,8 @@ module RGeo
       # Only available since GEOS 3.8+
       def make_valid
         @factory.wrap_fg_geom(@fg_geom.make_valid, nil)
+      rescue ::Geos::GEOSException
+        raise Error::UnsupportedOperation
       end if ::Geos::FFIGeos.respond_to?(:GEOSMakeValid_r)
 
       def equals?(rhs)
--- a/lib/rgeo/geos/ffi_feature_methods.rb
+++ b/lib/rgeo/geos/ffi_feature_methods.rb
@@ -97,11 +97,9 @@ module RGeo
       end
 
       def boundary
-        if self.class == FFIGeometryCollectionImpl
-          nil
-        else
-          @factory.wrap_fg_geom(@fg_geom.boundary, nil)
-        end
+        @factory.wrap_fg_geom(@fg_geom.boundary, nil)
+      rescue ::Geos::GEOSException
+        raise Error::InvalidGeometry, "Operation not supported by GeometryCollection"
       end
 
       def as_text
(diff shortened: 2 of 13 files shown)


test/common/validity_tests.rb
```diff
@@ -53,7 +53,7 @@ def test_validity_make_valid
           skip "#make_valid not handled by current implementation" unless implements_make_valid?
 
           assert_equal(bowtie_polygon_expected_area, bowtie_polygon.make_valid.area)
-          assert_raises(RGeo::Error::InvalidGeometry) do
+          assert_raises(RGeo::Error::UnsupportedOperation) do
```
keithdoggett, line 56: So to fix this for all do we need to check unsupported operation before validity?

doc/Factory-Compatibility.md
```diff
@@ -408,7 +408,7 @@ _Note: This list is not exhaustive of all the methods defined by each factory. T
 |                         `Collection#as_binary` |  ✅   |    ✅    |  ✅  |   ✅    |     ✅     |     ✅      |     ✅     |
 |                            `Collection#empty?` |  ✅   |    ✅    |  ✅  |   ✅    |     ✅     |     ✅      |     ✅     |
 |                           `Collection#simple?` |  ✅   |    ✅    |  ✅  |   ✅    |     ❌     |     ✅      |     ❌     |
-|                          `Collection#boundary` |  ✅   |    ❌    |  ✅  |   ❌    |     ❌     |     ✅      |     ❌     |
+|                          `Collection#boundary` |  ❌   |    ❌    |  ❌  |   ❌    |     ❌     |     ❌      |     ❌     |
```
BuonOmo, line 411: Geos would raise when calling this method, hence now we may consider removing this line altogether?

lib/rgeo/geos/ffi_feature_methods.rb
```diff
@@ -97,11 +97,9 @@ def envelope
...
 
       def boundary
-        if self.class == FFIGeometryCollectionImpl
-          nil
-        else
-          @factory.wrap_fg_geom(@fg_geom.boundary, nil)
-        end
+        @factory.wrap_fg_geom(@fg_geom.boundary, nil)
+      rescue ::Geos::GEOSException
+        raise Error::InvalidGeometry, "Operation not supported by GeometryCollection"
```
BuonOmo, line 102: We could raise `UnsupportedOperation` however this would make it inconsistant with the CAPI. And the CAPI workflow is not able to guess we're in that specific context without asserting even more from error strings.

doc/Factory-Compatibility.md
keithdoggett: Do you mean remove the method from `GeometryCollection` entirely? I'm ok with that, though it might make more sense to raise `UnsupportedOperation` instead.

lib/rgeo/geos/ffi_feature_methods.rb
keithdoggett: I think making it consistent with CAPI is better in this case.

keithdoggett (approved):
LGTM! I think we just have to figure out what to do with `#boundary` on a `GeometryCollection`, but this looks great.

doc/Factory-Compatibility.md
BuonOmo: I'm OK with that solution as well. I looked quickly what SFS 1.1 (https://gismanual.com/relational/99-049_OpenGIS_Simple_Features_Specification_For_SQL_Rev_1.1.pdf) has to say about it: it looks like a collection should have a boundary method, however the behaviour of such a method is not explained...

If you could have a second look on that, maybe you'll understand better

keithdoggett: Yeah, it seems unclear there. Seems like it's possible given a certain collection of geometries depending on the mix of them. Here's the GEOS implementation (https://github.com/libgeos/geos/blob/1dc4b60637c6d1c190c44accd2e524aef57489d7/src/geom/GeometryCollection.cpp#L198-L201) where they just throw an exception. I'd say it's probably better to align with GEOS and just raise `UnsupportedOperation` or something when we try to call it.

keithdoggett: I think leaving it as `InvalidGeometry` in the FFI and CAPI implementations is fine though.

BuonOmo: Ok then we're good to go :)

BuonOmo:
@keithdoggett there is only one issue left I think, it is to understand the various notices, and if any of those is a risk (`DEBUG=1 rake`).

Unfortunately, I don't think I'll have time to do it soon

image

keithdoggett:
@BuonOmo I ran the tests locally and looked through the notices. It looks like all of them are coming from the validity tests since GEOS will log a notice in some cases when a geometry is invalid.

```c
#include <geos_c.h>
#include <stdlib.h>
#include <stdio.h>
#include <stdarg.h>
#include <math.h>

void notice(const char *fmt, ...)
{
    va_list ap;
    fprintf(stdout, "NOTICE: ");
    va_start(ap, fmt);
    vfprintf(stdout, fmt, ap);
    va_end(ap);
    fprintf(stdout, "\n");
}

void log_and_exit(const char *fmt, ...)
{
    va_list ap;
    fprintf(stdout, "ERROR: ");
    va_start(ap, fmt);
    vfprintf(stdout, fmt, ap);
    va_end(ap);
    fprintf(stdout, "\n");
    exit(1);
}

int main()
{
    initGEOS(notice, log_and_exit);

    GEOSGeometry *gc = GEOSGeomFromWKT("POLYGON ((0 0, 1 1, 1 0, 0 1, 0 0))");
    GEOSisValid(gc);

   GEOSGeom_destroy(gc);
   finishGEOS();
   return 0;
}
```

The above will log 
```
NOTICE: Self-intersection at or near point 0.5 0.5
```
when run.

So those notices are fine and expected given that we are calling `GEOSisValid` on invalid geometries in some of our tests.

BuonOmo:
As discussed privately, every single of those notices come from validity related tests, moreover we don't really care about those warning (handling must be done otherwise, by throwing an error).

Hence the merge!

PR #287, opened 2022-02-09
Fix the raising of ParseError in the CoordSys::CS::WKTParser

mcls:
Currently, these call would have failed with:

  NoMethodError: undefined method `ParseError'

### Summary

I came across this minor issue while parsing some invalid WKT. The NoMethodError was confusing, fixing the call will raise the proper ParseError.


keithdoggett (approved):
LGTM! Good catch finding this

PR #290, opened 2022-02-10
Fix Psych::DisallowedClass errors in tests for ruby 3.1 / head

mcls:
Thanks for merging my previous PR so swiftly!  

I noticed some failures on the latest version of ruby in CI because of a Psych::DisallowedClass error. This is because Psych 4.0 aliases `load` to `safe_load`. (Related release notes: https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/)

Using `unsafe_load` and ensuring that the latest version of [Psych](https://github.com/ruby/psych) is used, regardless of the ruby version, fixes this. (I believe the Psych version works with all the tested ruby versions.)

### DIFF
--- a/rgeo.gemspec
+++ b/rgeo.gemspec
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "ffi-geos", "~> 1.2"
   spec.add_development_dependency "minitest", "~> 5.11"
+  spec.add_development_dependency "psych", ">= 4.0"
   spec.add_development_dependency "rake", "~> 13.0"
   spec.add_development_dependency "rake-compiler", "~> 1.0"
   spec.add_development_dependency "rubocop", "~> 1.8.1"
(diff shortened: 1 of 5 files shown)


keithdoggett:
Hi @mcls thanks for submitting another PR for this. In the `v3-dev` branch we have handled this already (https://github.com/rgeo/rgeo/pull/271/commits/0f0a7b3161aa0e277abe5e63c5f638f70f879b61). I'm fine with waiting for v3 to be merged in to handle it, but if it's causing a lot of issues we can just cherry-pick that commit. Thoughts @BuonOmo?

mcls:
Oops, didn't notice the v3 branch. No rush from my end.

BuonOmo:
Hey @mcls, thanks for the PR, I like your way of solving it.

@keithdoggett since tests will still be red whether we merge the PR or not, I'd say we don't have to.

@mcls sorry for bothering you with that work being done already :/ I can assure you there are plenty of tasks to do here, and we are grateful to see you contribute!

Issue #272, opened 2021-08-30
Ship GEOS with the project for MRI users


BuonOmo:
The [V3 roadmap](https://github.com/rgeo/rgeo/issues/247) includes our goal of having an easier installation process.

Since our pure ruby implementation is incomplete, and our the FFI version is way slower than the Geos CAPI version, it is clear that we want to keep the CAPI, and make it easier to install if we can.

Some idiomatic projects (such as https://github.com/sass/sassc-ruby) are showing us a good way to operate: intregrate GEOS directly in the CAPI build. There are tons of other projects that are doing so, you can just [search github for those](https://github.com/search?l=&o=asc&q=filename%3Aextconf.rb+language%3ARuby&s=indexed&type=Code).

Once we have a fixed version of GEOS shipped with RGeo, we also want one to be able to still be able to install its custom GEOS version.

One last thing, including GEOS version in our version, or a helper would be nice... Think `SELECT postgis_full_version()` for RGeo.

I am aware that this task may be a challenge for a first contributor. However, if you want to contribute, we'll be there to work together 💪

BuonOmo:
Note on advancement, there is #291 for binding another library than geos-config, and there is https://github.com/buonomo/test-geos for downloading and installing geos in extconf.rb

BuonOmo:
Current state:

1. we are still wondering how this would end up working
2. the dreamed state would be to ditch the pure ruby implementation in favor of the geos one, since it can come along
3. we like pg's extconf.rb and way to handle that
4. we may want to cache some pre-compiled versions at the rubygems level to have a quicker installation for users

eis-ioki:
I would love to see rgeo brings its own geos automatically, just like nokogiri brings its own libxml: https://nokogiri.org/#cruby

BuonOmo:
@eis-ioki thanks for the feedback! Do you want to contribute on that topic?

PR #291, opened 2022-02-25
Support GEOS Installation from Library

keithdoggett:
### Summary

Allows GEOS to be linked with a library (.so, .a, .dll, etc.) by specifying the `RGEO_GEOS_LIB_DIR` environment variable. 

### Other Information

This works, but there are still a couple issues that need to be discussed/resolved

- [ ] Due to how `find_library` works if `libgeos_c` is not found in the `RGEO_GEOS_LIB_DIR` directory and it is installed on the system (`/usr/local/lib` for example), it will silently fall back on the system installed version instead of raising an error.
- [ ] We may want to discuss automatically falling back on the `geos-config` flow if there is an error during the from library flow.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -18,7 +18,7 @@ task :clean do
   clean_files = %w[pkg tmp] +
     Dir.glob("ext/**/Makefile*") +
     Dir.glob("ext/**/*.{o,class,log,dSYM}") +
-    Dir.glob("**/*.{bundle,so,dll,rbc,jar}") +
+    Dir.glob("{ext,tmp,lib}/**/*.{bundle,so,dll,rbc,jar}")
     Dir.glob("**/.rbx")
 
   clean_files.each { |path| rm_rf path }
--- a/ext/geos_c_impl/extconf.rb
+++ b/ext/geos_c_impl/extconf.rb
@@ -6,14 +6,40 @@
 #
 # -----------------------------------------------------------------------------
 def create_dummy_makefile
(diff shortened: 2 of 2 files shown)


BuonOmo:
@keithdoggett that's nice thanks, indeed the fallback may be really missleading if someone is expecting the build from a specific library.

Also, and more importantly, I'm not sure where we want to go with this. E.g. in which circumstances someone would want to load geos from somewhere else? And in that case, why not loading geos from somewhere in the path of this library (https://github.com/rgeo/rgeo/issues/272).

Hence I think this is a nice step towards getting in that place, since we'll _only_ have to specify GEOS's local path then, but I just want to be sure about the reason why you started from that end :)

keithdoggett:
@BuonOmo yeah the main reason was to make that packaging easier eventually. I guess I figured that this was the hard part since I hadn't seen a C extension built in this way before, but there are some other use cases.

1) Some programs ship with GEOS and if you're using that library for some part of your system it may make sense to ensure that RGeo has the same version so you're not running different versions of GEOS. (one example: https://www.cockroachlabs.com/docs/stable/install-cockroachdb-linux.html).

2) Right now the only way to install it is if `geos-config` is installed on your system. In some environments it may be easier to work with the compiled libraries directly, especially if you want to use a version that is not available on the package manager by default. This just adds some more flexibility.

BuonOmo:
I definitely agree with you

> I guess I figured that this was the hard part since I hadn't seen a C extension built in this way before

I've seen some, for instance the ruby-sass!

PR #297, opened 2022-03-30
Only require geos/ffi_* if FFI supported

omarkhan:
### Summary

https://github.com/rgeo/rgeo/pull/275 introduced a hard dependency on `ffi-geos`: https://github.com/rgeo/rgeo/blob/fb0bfcdd0c3a1fdd67ab3f0ccff9cccb68d50336/lib/rgeo/geos/ffi_feature_methods.rb#L9

I suspect this was unintentional, since there is an alternative CAPI interface to GEOS that can be used without `ffi-geos`, and `ffi-geos` is not listed as a dependency in the gemspec.

This pull request adds a conditional to only require geos/ffi_* if `ffi-geos` is installed. This allows RGeo to be used without `ffi-geos`.

### Other Information

Thanks for all your hard work on v3! It's great work, much appreciated.

### DIFF
--- a/History.md
+++ b/History.md
@@ -24,6 +24,7 @@
 **Bug Fixes**
 
 * Fix CAPI equality (`==`) comparison on macos by using `GEOSEqualsExact_r` #270
+* Only require geos/ffi_* if FFI supported #297
 
 ### 2.4.0 / 2022-01-19
 
--- a/lib/rgeo/geos.rb
+++ b/lib/rgeo/geos.rb
@@ -33,9 +33,6 @@ module RGeo
       require_relative "geos/capi_feature_classes"
       require_relative "geos/capi_factory"
     end
-    require_relative "geos/ffi_feature_methods"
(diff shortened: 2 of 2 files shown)


keithdoggett (approved):
LGTM! Thank you very much for testing V3 and submitting a fix. It's greatly appreciated.

BuonOmo (approved):
LGTM! Good catch 👀

BuonOmo:
> @omarkhan looks like you may need to merge with `master` to get back up to date.
> 
> @BuonOmo I guess we should redirect the branch to `v3-dev` as well and include this in the next tranche of fixes?

Why not master directly ?

@omarkhan could you add a changelog entry ?

omarkhan:
@BuonOmo is this okay? https://github.com/rgeo/rgeo/pull/297/commits/e590521e0d3163d887d3f43cc7e8ead0bef98f75

PR #301, opened 2022-05-04
Add ruby_memcheck and fix leaks!

BuonOmo:
## Summary

The [ruby_memcheck](https://github.com/Shopify/ruby_memcheck) gem provides a hook to valgrind for someone who would like to check his library against valgrind memcheck. I've added it to rgeo, and fixed some memory leaks, thanks to @peterzhu2118 for his help and the gem!

There are two main fixes

1. use `ruby_xfree` against ruby allocator, rather than free, those may not be the same malloc/free (and not compatible).
2. fix a memory leak involving `#cast` ruby calls from the CAPI. See commit message for details.

## What's next

I'll open another PR that integrates memcheck to the CI, this is not that easy to integrate properly for now since there is no clear way to output potential errors, or filter out false positives. Will do so :)

## Mac users

If you are like me running on mac, and still want to use valgrind, here's the Dockerfile I used:

```dockerfile
FROM ruby:3.0

ARG work_dir=/ bundle_dir=/usr/local/bundle # if you want to mimic you PWD and .bundle

WORKDIR ${work_dir}
RUN bundle config set --local path ${bundle_dir}

RUN apt update && apt install -y libgeos-dev valgrind

COPY Gemfile rgeo.gemspec ./
COPY lib/rgeo/version.rb ./lib/rgeo/
RUN bundle install

COPY . .
RUN rake compile


CMD ["rake", "test:valgrind"]
```

### DIFF
--- a/ext/geos_c_impl/ruby_more.c
+++ b/ext/geos_c_impl/ruby_more.c
@@ -1,0 +1,65 @@
+/*
+	Utilities for the ruby CAPI
+*/
+
+#ifndef RGEO_GEOS_RUBY_MORE_INCLUDED
+#define RGEO_GEOS_RUBY_MORE_INCLUDED
+
+#include <ruby.h>
+
+#include "preface.h"
+#include "ruby_more.h"
+
+RGEO_BEGIN_C
+
+struct funcall_args
+{
+	VALUE recv;
+	ID mid;
+	int argc;
+	VALUE *argv;
+};
+
+static VALUE inner_funcall(VALUE args_)
+{
+	struct funcall_args *args = (struct funcall_args *)args_;
+	return rb_funcallv(
+		args->recv,
+		args->mid,
+		args->argc,
+		args->argv);
+}
+
+VALUE rb_protect_funcall(VALUE recv, ID mid, int *state, int n, ...)
+{
+	struct funcall_args args;
+	VALUE *argv;
+	va_list ar;
+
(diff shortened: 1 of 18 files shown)


ext/geos_c_impl/ruby_more.c
```diff
@@ -0,0 +1,65 @@
...
+
+static VALUE inner_funcall(VALUE args_)
+{
+	struct funcall_args *args = (struct funcall_args *)args_;
+	return rb_funcallv(
+		args->recv,
+		args->mid,
+		args->argc,
+		args->argv);
+}
```
keithdoggett, line 31: why do we need to break out the `inner_funcall` like this?

keithdoggett (approved):
LGTM just have one question

ext/geos_c_impl/ruby_more.c
BuonOmo: `rb_protect` take a function as its first argument, when you look at the source code, it is related to how ruby handles `longjmp`: it goes like:

1. prepare security in case of longjmp
2. call the actual fn
3. remove the security
4. finally return

I could have done all that in the `rb_protect_funcall` method but it would have meant copying a lot of ruby source code, and using internal macros!

To go further you could check https://blog.peterzhu.ca/ruby-c-ext-part-8/ and the source code of `rb_protect` and `rb_ensure` for instance

PR #302, opened 2022-05-04
Add New OGC SFS 1.2 Methods

### LINKED ISSUE #267: Update methods to match the current OGC SFS version (1.2)
Add the following methods to the `Geometry` class.

- [ ] `coordinate_dimension`
- [ ] `spatial_dimension`
- [ ] `3d?`
- [ ] `measured?`
- [ ] `locate_along`
- [ ] `locate_between`

keithdoggett:
Fixes #267 

I've implemented all the methods except for `location_along` and `locate_between`. I'm trying to figure out if there's a fairly simple way to do it. PostGIS just returns a `MultiPoint` object and doesn't worry about preserving the geometries if possible. If there's an easy way to flatten all the points of an arbitrary geometry it would be a simple implementation, but from my POV it seems like each geometry type would need a custom implementation for each factory type.

### DIFF
--- a/History.md
+++ b/History.md
@@ -20,6 +20,8 @@
 * Add `SweeplineIntersector` class to cartesian library
 * Add `PlanarGraph` class to cartesian library
 * Change `validate_geometry` methods to `init_geometry` methods that handle only quality of life functionality instead of validation as well
+* Add OGC SFS 1.2 methods to `RGeo::Feature::Geometry`.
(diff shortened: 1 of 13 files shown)


BuonOmo (approved):
I would not implement those now either: GEOS has not, lets wait for someone to ask for it (or even better, contribute it!)

Btw: don't forget history! (we should add that to the github template)

PR #304, opened 2022-05-05
Add memcheck to the CI

BuonOmo:


Rakefile
```diff
@@ -6,7 +6,11 @@ require "ruby_memcheck"
 require "bundler/gem_tasks"
 require "yard"
 
-RubyMemcheck.config(binary_name: "geos_c_impl")
+RubyMemcheck.config(
+  binary_name: "geos_c_impl",
+  # valgrind_generate_suppressions: true,
+  valgrind_suppressions_dir: "test/valgrind/suppressions"
+)
```
keithdoggett, line 13: So `RubyMemcheck` adds a `test:valgrind` task to our list of Rake commands and we just configure it here?

test/valgrind/suppressions/geos_c_impl_ruby.supp
```diff
@@ -0,0 +1,16 @@
...
+  ...
+  fun:main
+}
+{
+  <ruby_flo_to_s_false_positive>
+  Memcheck:Overlap
+  fun:__memcpy_chk
+  fun:memmove
+  fun:flo_to_s
+}
```
keithdoggett, line 16: Do we need this in version control or is this the output of a test?

test/valgrind/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+COPY Gemfile rgeo.gemspec ./
+COPY lib/rgeo/version.rb ./lib/rgeo/
+RUN bundle install
+
+COPY . .
+RUN rake compile
+
+
+CMD ["rake", "test:valgrind"]
```
keithdoggett, line 18: Is this just for convenience of testing? Or does RubyMemcheck specify that we use docker

keithdoggett (approved):
LGTM! Just left a few questions

Rakefile
BuonOmo: Exactly

test/valgrind/suppressions/geos_c_impl_ruby.supp
BuonOmo: We need this, it is a _suppression file_. Ruby has some memory issues itself because it doesn't clean up everything at exit. This file tells memcheck that those issues are to be ignored.

The second one by the way seems like a real ruby bug, but I didn't find a small reproducible case to open an issue in the ruby bug tracker..

test/valgrind/Dockerfile
BuonOmo: This is related to `bin/mac_valgrind`, which I could actually rename `bin/valgrind` maybe? This allows users to run the `rake test:valgrind` on a computer that cannot have valgrind

test/valgrind/suppressions/geos_c_impl_ruby.supp
keithdoggett: That makes sense thanks

test/valgrind/Dockerfile
keithdoggett: ahh yeah maybe we rename it to `bin/valgrind` or `bin/docker_valgrind`. But doesn't matter too much.

PR #310, opened 2022-06-07
feat: add invalid_reason_location method

aleksejleonov:
### Summary

This PR adds `invalid_reason_location` method (for GEOS-backed factories).

Example:
```ruby
# From validity tests:
hourglass = @factory.linear_ring([point1, point2, point3, point4, point1])
poly = @factory.polygon(hourglass)

poly.invalid_reason_location
# => #<RGeo::Geos::CAPIPointImpl:0xXXXXXX "POINT (0.5 0.5)">
```

- It's a nice to have feature.
- Helps migrate from RGeo 2 to RGeo 3 (as `invalid_reason` no longer returns the location of the invalid point).

### Other Information

GEOS has a very nice feature of the `GEOSisValidDetail_r` method to return the exact location that makes geometry invalid.

```C
# https://github.com/libgeos/geos/blob/b8331a9cb25dda85731ba10ae247d25f59723d10/capi/geos_ts_c.cpp#L758
char
    GEOSisValidDetail_r(GEOSContextHandle_t extHandle, const Geometry* g,
                        int flags, char** reason, Geometry** location)
```

Knowing the exact location helps a lot to fix invalid geometry.

For example, when you have 100_000 points GeoJSON polygon (created from some noisy GPS track), and there is `Self-intersection` somewhere, it's tough to find the exact place without `invalid_reason_location`.

In RGeo 2, we have used the `invalid_reason` method, parsing its text output with regexp (it contains point location as plain text).

In RGeo 3, `invalid_reason` became unified between Ruby and GEOS-backed factories (that's great 👍 ). But now, it returns only the name of the reason, without location. So, `invalid_reason_location` will fix this gap and helps RGeo2 to RGeo 3 migration.

### DIFF
--- a/History.md
+++ b/History.md
@@ -3,6 +3,7 @@
 **Minor Changes**
 
 * Implement `coordinate_dimension`, `spatial_dimension`, `is_3d?` and `measured?` for all factories.
+* Add `invalid_reason_location` method to the CAPI factory #310
 
 **Bug Fixes**
 
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -1058,6 +1058,32 @@ static VALUE method_geometry_invalid_reason(VALUE self)
   return result;
 }
 
+static VALUE method_geometry_invalid_reason_location(VALUE self)
+{
+  VALUE result;
+  RGeo_GeometryData* self_data;
+  const GEOSGeometry* self_geom;
+  GEOSGeometry* location = NULL;
+
+  result = Qnil;
+  self_data = RGEO_GEOMETRY_DATA_PTR(self);
+  self_geom = self_data->geom;
+  if (self_geom) {
+    // We use NULL there to tell GEOS that we don't care about the reason.
+    switch(GEOSisValidDetail_r(self_data->geos_context, self_geom, 0, NULL, &location)) {
+      case 0: // invalid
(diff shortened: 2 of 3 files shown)


keithdoggett (approved):
LGTM! Thanks @aleksejleonov this is definitely great to have

BuonOmo (approved):
Looks Great to me ! @aleksejleonov would you mind telling the world about your change in the `History.md` file ?

aleksejleonov:
> Looks Great to me ! @aleksejleonov would you mind telling the world about your change in the `History.md` file ?

Yes, I will update PR now 👍

aleksejleonov:
Tests failed due to some timeout in GitHub Actions CI:
https://github.com/rgeo/rgeo/runs/6804276614?check_suite_focus=true

```
Run sudo apt-add-repository ppa:ubuntugis/ubuntugis-unstable
 Unstable releases of Ubuntu GIS packages. These releases are more bleeding edge and while generally they should work well, they dont receive the same amount of quality assurance as our stable releases do.
 More info: https://launchpad.net/~ubuntugis/+archive/ubuntu/ubuntugis-unstable
Error: retrieving gpg key timed out.
Error: Process completed with exit code 1.
```

BuonOmo:
Ran the tests again, it works!

Thank you for your contribution @aleksejleonov, greatly appreciated

PR #313, opened 2022-06-24
feature: add polygonize method to CAPI factory

aleksejleonov:
### Summary

Adding GEOS `polygonize` to the CAPI factory.

```ruby
input = RGeo::Geos.factory.parse_wkt("LINESTRING(0 0, 1 1, 1 0, 0 0)")
input.polygonize.as_text

# GEOMETRYCOLLECTION(POLYGON ((0 0, 1 1, 1 0, 0 0)))
```

1. `polygonize` added as a common `geometry` method (not only for `Linestring`, `MultiLinestring`, `GeometryCollection`) to be aligned with GEOS API. It works predictably on `Point` and `Polygon` geometries, and this type of unified API seems logical (covered in tests).
2. This PR focuses only on the old `polygonize` GEOS method. There is a newer `polygonize_full` GEOS API method (starting GEOS 3.3) that returns not only a polygonized result, but also detailed error info about Dangles, Cut Edges, and Invalid Ring Lines.

### Other Information

GEOS `polygonize` description:
> Polygonizes a set of Geometries which contain linework that represents the edges of a planar graph.
> All types of Geometry are accepted as input; the constituent linework is extracted as the edges to be polygonized.
> The edges must be correctly noded; that is, they must only meet at their endpoints and not overlap anywhere. If your edges are not already noded, run them through [GEOSUnaryUnion()](https://libgeos.org/doxygen/geos__c_8h.html#ada5b7b39a81b722b97371736ff0b8d29) first. Polygonization will accept incorrectly noded input but will not form polygons from non-noded edges, and reports them as errors.
>
> https://libgeos.org/doxygen/geos__c_8h.html#a9d98e448d3b846d591c726d1c0000d25

This PR is based on @tyfoan work and all comments in #222.

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,7 @@
 
 * Implement `coordinate_dimension`, `spatial_dimension`, `is_3d?` and `measured?` for all factories.
 * Add `invalid_reason_location` method to the CAPI factory #310
+* Add `polygonize` method to the CAPI factory (@aleksejleonov, @tyfoan) #313
 
 **Bug Fixes**
 
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -1121,6 +1121,43 @@ static VALUE method_geometry_point_on_surface(VALUE self)
   return result;
 }
 
+/**
+  * call-seq:
+  *   some.polygonize -> RGeo::Feature::GeometryCollection
+  *
+  * Polygonizes a set of Geometries which contain linework that
+  * represents the edges of a planar graph.
+  *
+  * All types of Geometry are accepted as input;
+  * the constituent linework is extracted as the edges to be polygonized.
+  * 
+  * The edges must be correctly noded;
+  * that is, they must only meet at their endpoints and not overlap anywhere.
+  *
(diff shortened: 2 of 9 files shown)


aleksejleonov:
@keithdoggett, sorry for pushing on this 🙇‍♂️ 

I have one case uncovered by tests in this PR.

By documentation of GEOS `polygonize`, it could return `NULL` in case of exception.
But with any type of invalid geometry, I tried on input, it always returns a valid result `GEOMETRYCOLLECTION EMPTY`.

So, next lines are untested:
```C
if (geos_polygon_collection == NULL) {
  rb_raise(rb_eGeosError, "GEOS can't polygonize this geometry.");
}
```

keithdoggett (approved):
@aleksejleonov thank you for implementing another feature! Everything looks clear in the code and your tests are robust and simple so this looks great to me.

To your point about the null case not being tested, I think that's fine. It's fairly straightforward that it would raise an error in that case and we're not doing anything else so no worries. I've also had issues in the past with GEOS where I couldn't produce a `NULL` result even though the docs stated that it should have.

BuonOmo (approved):
LGTM as well, thanks for the help ! I'd maybe mention @tyfoan in the history, what do you think ?

aleksejleonov:
@BuonOmo 

I have added mention about @tyfoan into `History.md`.

PR #314, opened 2022-07-07
refactor: Remove `SRSDatabase` and `factory_generator` methods

BuonOmo:
There is no open source usage of SRSDatabase anywhere, see:

- https://grep.app/search?q=srs_database&filter[lang][0]=Ruby
- https://grep.app/search?q=SRSDatabase&filter[lang][0]=Ruby

As we'd like to simplify the codebase, we'll remove this for now, if
one wants it back, we're opened to extract this into a separate gem.

The only purpose of `factory_generator` methods was related to that, and
could simply be implemented using a `proc`, hence we also remove those
from the interface.

### Summary

Provide a general description of the code changes in your pull
request... were there any bugs you had fixed? If so, mention them. If
these bugs have open GitHub issues, be sure to tag them here as well,
to keep the conversation linked together.

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include benchmarks.

Thanks for contributing!


keithdoggett (approved):
LGTM! I'll just note that this is part of a bigger coordinate system/proj refactor that is upcoming.

Issue #317, opened 2022-07-07
Proj Rework: Clean up Core Gem


keithdoggett:
As part of a proj and coord_sys refactoring project:

We need to remove references to `proj4` in the core gem so that they can be properly decoupled.

- [x] `SimpleMercatorProjector`: Subclass `CoordinateTransform` and implement those methods.
- [x] Factory instantiation: Rely on the `srid` keyword to define the crs, or a new `coord_sys` parameter that accepts a CoordSys object. We could have a combination where if SRID is given, but not `coord_sys`, we try to make a `coord_sys` with an available adapter.
- [x] Proj4Projector: Likely remove, but we could create a new Projector class that subclasses CT, although it probably makes more sense for the projector param on geographic factories to just take a CT and ensure that the source_cs is 4326.
- [x] SRSDatabase Module: Remove.
- [x] `CoordSys#supported?` & `#check!`: Remove.
- [x] `Feature#cast`: Use the `CoordSys` information from each factory to create a CT that will transform the geometry.

PR #322, opened 2022-09-05
Proj Rework: Factory Interface Changes

keithdoggett:
### Summary

Removes the `proj4` option from all factories and relies on the `coord_sys` option/attribute to act in its place.

The goal is to decouple this gem from the rgeo-proj4 gem. By replacing `proj4` with `coord_sys` all we will need is an implementation of `CoordSys::CS::CoordinateSystem` in each factory as `@coord_sys` and if a `CoordinateTransform` can be created from them, a transform can be done between factories. 

This change allows us to implement any coordinate transform adapter we desire as long as it provides an implementation of `CoordSys::CS::CoordinateSystem` and `CoordSys::CS::CoordinateTransform`. More importantly, it allows us to use the proj4 gem as an adapter that plugs into this system rather than requiring the core gem to check if proj4 is available anytime a transform is to be done.

Changes:

- Remove `proj4` option from all factory creation interfaces
- Remove `proj4` attribute and references in all factory methods.
- Replace `proj4` references with `coord_sys` where appropriate
- Add `CoordSys::CONFIG.default_coord_sys_class` that specifies which `CoordSys::CS::CoordinateSystem` implementation should be the default one when trying to make `coord_sys` objects.
- Automatically try to create `coord_sys` objects from the given `srid` or `coord_sys` string based on the `default_coord_sys_class`.
- Add `coord_sys_class` option to all factories that can be used to override the `default_coord_sys_class` with a specific implementation.

### Other Information

This is not functional currently and more PRs will have to be done to support transforms again, but this is the first step. See: #317

### DIFF
--- a/lib/rgeo/coord_sys.rb
+++ b/lib/rgeo/coord_sys.rb
@@ -6,6 +6,7 @@
 #
 # -----------------------------------------------------------------------------
 
+require 'ostruct'
 require_relative "coord_sys/cs/factories"
 require_relative "coord_sys/cs/entities"
 require_relative "coord_sys/cs/wkt_parser"
@@ -24,6 +25,9 @@ module RGeo
   # systems, and other related concepts, as well as a parser for the WKT
   # format for specifying coordinate systems.
   module CoordSys
+    CONFIG = OpenStruct.new
+    CONFIG.default_coord_sys_class = CS::CoordinateSystem
+
     # The only valid key is :proj4
     def self.supported?(key)
       raise(Error::UnsupportedOperation, "Invalid key. The only valid key is :proj4.") unless key == :proj4
--- a/lib/rgeo/coord_sys/cs/entities.rb
+++ b/lib/rgeo/coord_sys/cs/entities.rb
@@ -939,6 +939,15 @@ module RGeo
         def get_units(dimension)
           nil
         end
+
+        class << self
+          def create(defn, *optional)
+            # TODO: should this raise?
+            # Need this so we can maintain consistency with actual
+            # CoordinateSystem implementations
+            nil
+          end
+        end
       end
 
       # == OGC spec description
--- a/lib/rgeo/geographic/factory.rb
+++ b/lib/rgeo/geographic/factory.rb
@@ -36,17 +36,25 @@ module RGeo
         @coordinate_dimension += 1 if @support_m
         @spatial_dimension = @support_z ? 3 : 2
 
-        @srid = (opts[:srid] || 4326).to_i
-        @proj4 = opts[:proj4]
-        if @proj4 && CoordSys.check!(:proj4)
-          if @proj4.is_a?(String) || @proj4.is_a?(Hash)
-            @proj4 = CoordSys::Proj4.create(@proj4)
-          end
-        end
         @coord_sys = opts[:coord_sys]
-        if @coord_sys.is_a?(String)
-          @coord_sys = CoordSys::CS.create_from_wkt(@coord_sys)
+        coord_sys_class = opts[:coord_sys_class]
+        unless coord_sys_class.is_a?(Class)
+          coord_sys_class = CoordSys::CONFIG.default_coord_sys_class
         end
+
+        if @coord_sys.is_a?(String)
+          @coord_sys = coord_sys_class.create_from_wkt(@coord_sys)
+        end
+
+        unless @coord_sys.is_a?(CoordSys::CS::CoordinateSystem)
+          # TODO: should we raise here?
+          @coord_sys = nil
+        end
+        @srid = (opts[:srid] || 4326).to_i
+        if @coord_sys.nil? && srid != 0
+          @coord_sys = coord_sys_class.create(srid)
+        end
+
         @buffer_resolution = opts[:buffer_resolution].to_i
         @buffer_resolution = 1 if @buffer_resolution < 1
 
@@ -89,14 +97,14 @@ module RGeo
           @impl_prefix == rhs_.instance_variable_get(:@impl_prefix) &&
           @support_z == rhs_.instance_variable_get(:@support_z) &&
           @support_m == rhs_.instance_variable_get(:@support_m) &&
-          @proj4 == rhs_.instance_variable_get(:@proj4)
+          @coord_sys == rhs_.instance_variable_get(:@coord_sys)
       end
       alias == eql?
 
       # Standard hash code
 
       def hash
-        @hash ||= [@impl_prefix, @support_z, @support_m, @proj4].hash
+        @hash ||= [@impl_prefix, @support_z, @support_m, @coord_sys].hash
       end
 
       # Marshal support
@@ -113,7 +121,6 @@ module RGeo
           "wkbp" => @wkb_parser.properties,
           "bufr" => @buffer_resolution
         }
-        hash_["proj4"] = @proj4.marshal_dump if @proj4
         hash_["cs"] = @coord_sys.to_wkt if @coord_sys
         if @projector
           hash_["prjc"] = @projector.class.name.sub(/.*::/, "")
@@ -123,12 +130,6 @@ module RGeo
       end
 
       def marshal_load(data_) # :nodoc:
-        if (proj4_data = data_["proj4"]) && CoordSys.check!(:proj4)
-          proj4 = CoordSys::Proj4.allocate
-          proj4.marshal_load(proj4_data)
-        else
-          proj4 = nil
-        end
         if (coord_sys_data = data_["cs"])
           coord_sys = CoordSys::CS.create_from_wkt(coord_sys_data)
         else
@@ -143,7 +144,6 @@ module RGeo
           wkt_parser: symbolize_hash(data_["wktp"]),
           wkb_parser: symbolize_hash(data_["wkbp"]),
           buffer_resolution: data_["bufr"],
-          proj4: proj4,
           coord_sys: coord_sys
         )
         if (proj_klass = data_["prjc"]) && (proj_factory = data_["prjf"])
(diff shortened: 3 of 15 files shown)


lib/rgeo/coord_sys.rb
```diff
@@ -24,6 +25,9 @@ module RGeo
   # systems, and other related concepts, as well as a parser for the WKT
   # format for specifying coordinate systems.
   module CoordSys
+    CONFIG = OpenStruct.new
```
BuonOmo, line 28: I think `Struct` is better perf-wise:

```ruby
# frozen_string_literal: true

require "ostruct"
require "benchmark"
require "benchmark/ips"

def h1(str)
  puts
  puts str.upcase
  puts ?= * str.size
  puts
end

STRUCT = Struct.new(:foo, keyword_init: true)

h1 "creation"

Benchmark.ips do |x|
  x.report("Struct") { STRUCT.new(foo: "bar") }
  x.report("OpenStruct") { OpenStruct.new(foo: "bar") }
  x.compare!
end

h1 "record access"

struct = STRUCT.new(foo: "bar")
ostruct = OpenStruct.new(foo: "bar")

Benchmark.ips do |x|
  x.report("Struct") { struct.foo }
  x.report("OpenStruct") { ostruct.foo }
  x.compare!
end
```

```
CREATION
========

              Struct   533.208k i/100ms
          OpenStruct    27.978k i/100ms
              Struct      5.369M (± 2.1%) i/s -     27.194M in   5.067529s
          OpenStruct    281.461k (± 2.2%) i/s -      1.427M in   5.072052s

Comparison:
              Struct:  5368589.4 i/s
          OpenStruct:   281460.9 i/s - 19.07x  (± 0.00) slower


RECORD ACCESS
=============

              Struct     2.084M i/100ms
          OpenStruct   970.030k i/100ms
              Struct     21.106M (± 0.5%) i/s -    106.283M in   5.035739s
          OpenStruct      9.748M (± 0.6%) i/s -     49.472M in   5.075373s

Comparison:
              Struct: 21106238.1 i/s
          OpenStruct:  9747667.8 i/s - 2.17x  (± 0.00) slower
```

Although, accessing the value is not that slower in the end. However, not requiring ostruct is also a benefit, don't you think?

```suggestion
    CONFIG = Struct.new(:default_coord_sys_class).new(CS::CoordinateSystem)
```

IDK if this would be enough ?

lib/rgeo/geographic/factory.rb
```diff
@@ -36,17 +36,25 @@ def initialize(impl_prefix, opts = {}) # :nodoc:
...
-        @srid = (opts[:srid] || 4326).to_i
-        @proj4 = opts[:proj4]
-        if @proj4 && CoordSys.check!(:proj4)
-          if @proj4.is_a?(String) || @proj4.is_a?(Hash)
-            @proj4 = CoordSys::Proj4.create(@proj4)
-          end
-        end
         @coord_sys = opts[:coord_sys]
+        coord_sys_class = opts[:coord_sys_class]
+        unless coord_sys_class.is_a?(Class)
```
BuonOmo, line 41:
```suggestion
        unless coord_sys_class.respond_to?(:create_from_wkt)
```

more flexible if someone wants to wrap the class ?

lib/rgeo/geographic/factory.rb
```diff
@@ -36,17 +36,25 @@ def initialize(impl_prefix, opts = {}) # :nodoc:
...
         if @coord_sys.is_a?(String)
-          @coord_sys = CoordSys::CS.create_from_wkt(@coord_sys)
+          @coord_sys = coord_sys_class.create_from_wkt(@coord_sys)
+        end
+
+        unless @coord_sys.is_a?(CoordSys::CS::CoordinateSystem)
+          # TODO: should we raise here?
+          @coord_sys = nil
+        end
+        @srid = (opts[:srid] || 4326).to_i
```
BuonOmo, line 53:
```suggestion
        @srid = opts.fetch(:srid, 4326).to_i
```

lib/rgeo/coord_sys/cs/entities.rb
```diff
@@ -939,6 +939,15 @@ def get_axis(dimension)
         def get_units(dimension)
           nil
         end
+
+        class << self
+          def create(defn, *optional)
+            # TODO: should this raise?
```
BuonOmo, line 945: I guess raising is the consistent behaviour nowadays, what bothers you?

lib/rgeo/geographic/interface.rb
```diff
@@ -315,29 +299,23 @@ def simple_mercator_factory(opts = {})
...
       def projected_factory(opts = {})
-        CoordSys.check!(:proj4)
         if (projection_factory = opts[:projection_factory])
           # Get the projection coordinate systems from the given factory
-          projection_proj4 = projection_factory.proj4
-          unless projection_proj4
-            raise ArgumentError, "The :projection_factory does not have a proj4."
-          end
           projection_coord_sys = projection_factory.coord_sys
+          # TODO: may have to change to just CoordinateSystem because the proj4 implementation doesn't subclass ProjectedCoordinateSystem.
```
BuonOmo, line 305: Not sure here

lib/rgeo/geographic/interface.rb
```diff
@@ -348,47 +326,43 @@ def projected_factory(opts = {})
...
-          # Check the projection coordinate systems, and parse if needed.
-          if projection_proj4.is_a?(String) || projection_proj4.is_a?(Hash)
-            actual_projection_proj4 = CoordSys::Proj4.create(projection_proj4)
-            unless actual_projection_proj4
-              raise ArgumentError, "Bad proj4 syntax: #{projection_proj4.inspect}"
-            end
-            projection_proj4 = actual_projection_proj4
+
+          if projection_coord_sys.nil? && projection_srid != 0
+            projection_coord_sys = coord_sys_class.create(projection_srid)
```
BuonOmo, line 341: could we somehow refactor all of these to avoid repeating

lib/rgeo/geographic/interface.rb
```diff
@@ -403,30 +377,16 @@ def projected_factory(opts = {})
...
-        unless defined?(@proj44055)
-          @proj44055 = CoordSys.supported?(:proj4) && CoordSys::Proj4.create("+proj=longlat +a=6378137 +b=6378137 +towgs84=0,0,0,0,0,0,0 +no_defs")
-        end
-        @proj44055
-      end
-
       def coord_sys_4055
         unless defined?(@coord_sys_4055)
-          @coord_sys_4055 = CoordSys::CS.create_from_wkt('GEOGCS["Popular Visualisation CRS",DATUM["Popular_Visualisation_Datum",SPHEROID["Popular Visualisation Sphere",6378137,0,AUTHORITY["EPSG","7059"]],TOWGS84[0,0,0,0,0,0,0],AUTHORITY["EPSG","6055"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.01745329251994328,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4055"]]')
+          @coord_sys_4055 = CoordSys::CONFIG.default_coord_sys_class.create(4055)
```
BuonOmo, line 382: what is the point of these methods? Faster initialisation ?

BuonOmo:
This looks really good! I made some comments, I think some of them should be discussed now, but we're close to merging it

lib/rgeo/geographic/interface.rb
keithdoggett: The `Geographic` module contains some factories that allow you to project from 4326 or 4055 to another coord sys, so these just contain the definitions for the source coord systems of the factories.

lib/rgeo/geographic/interface.rb
keithdoggett: I think so. Proj4 just implements one class that can be any type of coordinate system (projected or geographic). But it has methods like `geographic?` that can be used to tell what type it is.

We should just add `geographic?` and `projected?` to the base `CS::CoordianteSystem` definition and children so we can rely on those.

lib/rgeo/coord_sys/cs/entities.rb
keithdoggett: Yeah I think the issue with this and the above comment is that if you're not using proj and just want to use a factory, then it would raise on instantiation if you specify an srid.

ex. RGeo::Geos.factory(srid: 4326) would raise with an `InvalidCoordinateSystem` error or something. Obviously not ideal, so maybe we should raise here, but rescue and assign null in the factory creation logic.

Perhaps we could actually give it a coordinate system but it raises once you try to transform because there's not valid transformation? That might actually be the best bet.

lib/rgeo/geographic/factory.rb
```diff
@@ -36,17 +36,25 @@ def initialize(impl_prefix, opts = {}) # :nodoc:
...
+
         if @coord_sys.is_a?(String)
-          @coord_sys = CoordSys::CS.create_from_wkt(@coord_sys)
+          @coord_sys = coord_sys_class.create_from_wkt(@coord_sys)
+        end
+
+        unless @coord_sys.is_a?(CoordSys::CS::CoordinateSystem)
+          # TODO: should we raise here?
+          @coord_sys = nil
+        end
```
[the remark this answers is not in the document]
keithdoggett: yeah I guess I can just keep it and if it's not a CoordinateSystem, it will end up raising down the line when you try to invoke a transformation on it

keithdoggett:
@BuonOmo I made a few changes now so we will always create a coord_sys if an SRID is given, it will just be the abstract base class that you can't do anything with.

I had to implement some form of WKT parsing for that class as well. It's not correct, but a lot of tests would fail if a CoordinateSystem couldn't be created from the WKTParser.

BuonOmo (approved):
I think the compromise is worth the benefits, LGTM

keithdoggett:
@BuonOmo FYI I made some tweaks to the WKT generations so it's now valid just slightly incorrect WKT, but that doesn't matter since it's an abstract class anyways.

PR #323, opened 2022-09-16
chore: Improve C ext debugging, format C code, update ruby code formatting

BuonOmo:
A first step before we add a formatting tool (I'm considering clang-format)! 

The maintainer mode is inspired by https://github.com/ged/ruby-pg/blob/master/ext/extconf.rb


keithdoggett (approved):
LGTM!

Issue #307, opened 2022-05-20
Concurrency problems with single app-wide factory?


durkie:
Hi RGeo folks -- I recently switched to declaring a single RGeo factory in `config/initializers/rgeo.rb` as recommended in the docs, along the lines of simply `RF = RGeo::Geos.factory(srid: 4326)`, and then I reference `RF` in various models/services throughout my app.

Around the same time, I noticed that many of my Sidekiq jobs seem to be failing intermittently with strange parsing errors, and I just put it together that this seems like it might be related to having a single, app-wide factory:
```
sidekiq_1      | 2022-05-20T01:06:57.208Z pid=1 tid=gp7mdij15 WARN: NoMethodError: undefined method `scan_until' for nil:NilClass
sidekiq_1      | 2022-05-20T01:06:57.210Z pid=1 tid=gp7mdij15 WARN: /usr/src/app/vendor/bundle/ruby/2.6.0/gems/rgeo-2.1.1/lib/rgeo/wkrep/wkt_parser.rb:396:in `next_token'
sidekiq_1      | /usr/src/app/vendor/bundle/ruby/2.6.0/gems/rgeo-2.1.1/lib/rgeo/wkrep/wkt_parser.rb:223:in `parse_coords'
sidekiq_1      | /usr/src/app/vendor/bundle/ruby/2.6.0/gems/rgeo-2.1.1/lib/rgeo/wkrep/wkt_parser.rb:267:in `parse_point'
sidekiq_1      | /usr/src/app/vendor/bundle/ruby/2.6.0/gems/rgeo-2.1.1/lib/rgeo/wkrep/wkt_parser.rb:202:in `parse_type_tag'
sidekiq_1      | /usr/src/app/vendor/bundle/ruby/2.6.0/gems/rgeo-2.1.1/lib/rgeo/wkrep/wkt_parser.rb:133:in `parse'
sidekiq_1      | /usr/src/app/vendor/bundle/ruby/2.6.0/gems/rgeo-2.1.1/lib/rgeo/geos/capi_factory.rb:288:in `parse_wkt'
sidekiq_1      | /usr/src/app/app/lib/osm.rb:132:in `block in get_sqlite_streets'
```

Looking at `wkt_parser.rb:396`, it looks like the error is that `@scanner` is nil, which seems like it can only happen if some other job called `clean_scanner` (`wkt_parser.rb:384`) at just the right time and set `@scanner` to nil right as my current job needed `@scanner.scan_until()`

[...]

PR #326, opened 2022-09-20
feat: Thread-safe `RGeo::WKRep::*` family

BuonOmo:
Fixes https://github.com/rgeo/rgeo/issues/307


.github/workflows/CI.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           sudo apt-get install -yqq libgeos-dev
       - name: Install Geos (Mac)
         if: matrix.os == 'macos'
-        run: brew install geos
+        run: HOMEBREW_NO_INSTALL_CLEANUP=1 brew install geos
```
keithdoggett, line 44: what's the need for this?

keithdoggett (approved):
This is very cool thanks for looking into this. LGTM!

.github/workflows/CI.yml
BuonOmo: We don't need to cleanup since we'll get rid of the instance later on, and that's one step less in the CI :)

PR #327, opened 2022-09-20
Pros Rework: `Feature#cast` Refactor

keithdoggett:
…. Add transform_coords to CoordinateSystem instance.

### Summary

* Use `coord_sys` instead of `proj4` to transform coordinates in the `cast` method.
* Remove `check!` and `supported?` from `CoordSys` module.
* Add `transform_coords` method to `CS::CoordinateSystem` instances.

### Other Information

Part of the larger proj rework. After this we really just have to refactor the mercator implementation and decide what to do with the `Proj4Projector`. We will also have to add a few methods to the `Proj4` coordinate system implementation in the rgeo-proj4 package as well.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -38,3 +38,33 @@ end
 if defined?(GC.auto_compact) == "method"
   GC.auto_compact = true
 end
+
+# Basic test class where transformations will translate
+# based on difference between "value" attribute
+class TestAffineCoordinateSystem < RGeo::CoordSys::CS::CoordinateSystem
+  def initialize(value, dimension, *optional)
+    super(value, dimension, *optional)
+    @value = value
+  end
+  attr_accessor :value
+
+  def transform_coords(target_cs, x, y, z = nil)
+    ct = TestAffineCoordinateTransform.create(self, target_cs)
+    ct.transform_coords(x, y, z)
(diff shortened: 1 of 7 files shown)


test/test_helper.rb
```diff
@@ -38,3 +38,33 @@ def psych_load(*args)
...
+end
+
+class TestAffineCoordinateTransform < RGeo::CoordSys::CS::CoordinateTransform
+  def transform_coords(x, y, z = nil)
+    diff = target_cs.value - source_cs.value
+    coords = [x + diff, y + diff]
+    coords << (z + diff) if z
+    coords
+  end
+end
```
BuonOmo, line 70: I think since this is used only in one file, I'd be more in favor of putting it directly in that file.

Or, we could have a `test/support` folder for those kinds of helpers, where we could also put the `data.wk{b,t}` that I made in my last PR

BuonOmo (approved):
One test organisation tidbit, but this can definitely be handled in a latter PR depending on your feedback.

Thanks for the great work, we can really see it shaping out nicely now!

test/test_helper.rb
keithdoggett: Agreed. Moved it into the types_test file.

PR #328, opened 2022-09-21
change interface of *Projector classes to rely on coord_sys

keithdoggett:
### Summary

Change `Proj4Projector` and `SimpleMercatorProjector` to use `coord_sys` objects instead of `proj4`.


keithdoggett:
@BuonOmo should we rename `Proj4Projector`? Maybe `GeographyProjector` or `CoordSysProjector`. Not sure what's best.

BuonOmo:
@keithdoggett I think we should indeed, I guess both are ok, even simply `Projector` could work since you're already under the module `Geography`

PR #329, opened 2022-09-21
V3 Proj Rework

keithdoggett:
### Summary

Removes references to `proj4` and the `rgeo-proj4` gem. Relies on `CoordSys::CS::CoordinateSystem` and `CoordSys::CS::CoordinateTransform` implementations to handle transformations.

Adds some various convenience methods and helpers to more easily assign `coord_sys` objects in factories.

Adds `coord_sys_class` option to define what implementation will be used to attempt to create a `CoordinateSystem`.

Adds `CoordSys::CONFIG.default_coord_sys_class` option to default to a specific `CoordinateSystem` implementation.

### Other Information

Some updates need to be done to the `rgeo-proj4` gem to make it compatible with this version.


keithdoggett:
Fixes #317

keithdoggett:
@BuonOmo this is just the combination of all the proj related PRs. I also fixed the history to reflect what we've changed since the last rc release.

lib/rgeo/coord_sys/cs/entities.rb
```diff
@@ -939,6 +939,51 @@ def get_axis(dimension)
...
+
+        # Not an OGC method, but useful for being able to
+        # transform directly from a CoordinateSystem object.
+        def transform_coords(target_cs, x, y, z = nil)
+          ct = CoordinateTransform.create(self, target_cs)
+          ct.transform_coords(x, y, z)
+        end
+
+        class << self
+          def create(defn, dimension = 2, *optional)
```
BuonOmo, line 963: What is the point of this undocumented `optional` arg list?

lib/rgeo/coord_sys/cs/wkt_parser.rb
```diff
@@ -79,6 +79,13 @@ def parse(containing_type = nil) # :nodoc:
             obj = VerticalDatum.create(args.shift(QuotedString), args.shift(Numeric), *args.create_optionals)
           when "LOCAL_DATUM"
             obj = LocalDatum.create(args.shift(QuotedString), args.shift(Numeric), *args.create_optionals)
+          when "CS"
```
BuonOmo, line 82: The only issue is: what happens when we parse wkt using geos and its default implementation ?

keithdoggett: Would the geos parser try to parse the CS stuff as well? I can check again, but I thought this parser is only used in `CS::*` modules and wouldn't be used when a geos factory is attempting to parse WKT geometries for example.

BuonOmo: Oh maybe not, you are right

BuonOmo (approved):
LGTM!

lib/rgeo/coord_sys/cs/entities.rb
keithdoggett: Most of the classes in this module inherit from the `Info` class either directly or through a chain of classes.

The `optional` arg is all of the optional parameters from the `Info` constructor

https://github.com/rgeo/rgeo/blob/0ca3e09321ab91e54fe39aa86b87a73a7c397eb5/lib/rgeo/coord_sys/cs/entities.rb#L432-L441

Most of those allow us to define tags and other optional pieces of data within other WKT blocks.

For example, we would define the `AUTHORITY` tag in the following WKT with the `authority_name` and `authority_code` optional args for each block that has an `AUTHORITY` tag. 
```
PROJCS["RGF93 v2 / CC42",
    GEOGCS["RGF93 v2",
        DATUM["Reseau_Geodesique_Francais_1993_v2",
            SPHEROID["GRS 1980",6378137,298.257222101],
            TOWGS84[0,0,0,0,0,0,0]],
        PRIMEM["Greenwich",0,
            AUTHORITY["EPSG","8901"]],
        UNIT["degree",0.0174532925199433,
            AUTHORITY["EPSG","9122"]],
        AUTHORITY["EPSG","9777"]],
    PROJECTION["Lambert_Conformal_Conic_2SP"],
    PARAMETER["latitude_of_origin",42],
    PARAMETER["central_meridian",3],
    PARAMETER["standard_parallel_1",41.25],
    PARAMETER["standard_parallel_2",42.75],
    PARAMETER["false_easting",1700000],
    PARAMETER["false_northing",1200000],
    UNIT["metre",1,
        AUTHORITY["EPSG","9001"]],
    AXIS["Easting",EAST],
    AXIS["Northing",NORTH],
    AUTHORITY["EPSG","9822"]]
```

keithdoggett: we could probably use kwargs instead so it's a little clearer what is actually being defined

PR #330, opened 2022-09-23
One factory

BuonOmo:
We could use only one geos handle since we are running in ruby, and our threads are protected by the GIL. 

This doesn't really improve performances though:

```bash
$ hyperfine \
	--prepare 'git checkout {tag}; rake compile' \
	--parameter-list tag main,one-factory \
	'TEST="test/geos_capi/*test.rb" rake test # {tag}'

Benchmark 1: TEST="test/geos_capi/*test.rb" rake test # main
  Time (mean ± σ):     894.6 ms ±  19.4 ms    [User: 623.4 ms, System: 201.7 ms]
  Range (min … max):   879.4 ms … 947.1 ms    10 runs
 
Benchmark 2: TEST="test/geos_capi/*test.rb" rake test # one-factory
  Time (mean ± σ):     894.0 ms ±  19.4 ms    [User: 629.0 ms, System: 197.8 ms]
  Range (min … max):   870.6 ms … 932.7 ms    10 runs
 
Summary
  'TEST="test/geos_capi/*test.rb" rake test # one-factory' ran
    1.00 ± 0.03 times faster than 'TEST="test/geos_capi/*test.rb" rake test # main'
```

### DIFF
--- a/ext/geos_c_impl/globals.c
+++ b/ext/geos_c_impl/globals.c
@@ -13,8 +13,6 @@
 
 RGEO_BEGIN_C
 
-GEOSContextHandle_t geos_context;
-
 VALUE rgeo_module;
 
 VALUE rgeo_feature_module;
@@ -45,18 +43,18 @@ VALUE rgeo_geos_multi_polygon_class;
 // GEOSIsValid_r (check for NOTICE_MESSAGE in GEOS codebase).
 // We still set it to make sure we do not miss any implementation
 // change. Use `DEBUG=1 rake` to show notice.
-#ifdef DEBUG
 static void
 notice_handler(const char* fmt, ...)
 {
+#ifdef DEBUG
   va_list args;
   va_start(args, fmt);
   fprintf(stderr, "GEOS Notice -- ");
   vfprintf(stderr, fmt, args);
   fprintf(stderr, "\n");
   va_end(args);
-}
 #endif
+}
 
 static void
 error_handler(const char* fmt, ...)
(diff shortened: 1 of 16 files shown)


ext/geos_c_impl/globals.c
```diff
@@ -34,9 +39,59 @@ VALUE rgeo_geos_multi_point_class;
...
+  } else {
+    rb_raise(rb_eGeosError, "%s", geos_error);
+  }
+}
+
 void
 rgeo_init_geos_globals()
 {
+  initGEOS(notice_handler, error_handler);
+
```
keithdoggett, line 94: Feels better handling this here rather than in the factory definitions

keithdoggett (approved):
LGTM

BuonOmo:
FYI @keithdoggett here's how I tested that it still behaves correctly after forking:

```c
  // globals.c, after initGEOS call
  VALUE rb_mProcess = rb_define_module("Process");

  GEOSWKTReader* reader = GEOSWKTReader_create();
    GEOSGeometry* geom = GEOSWKTReader_read(reader, "POINT(1 1)");
  GEOSSetSRID(geom, 1);

  for (int i = 0; i < 100; i++) {
    VALUE pid = rb_funcall(rb_mProcess, rb_intern("fork"), 0);
    if (pid != Qnil) continue;

    GEOSSetSRID(geom, i + 100);
    rb_funcall(rb_mProcess, rb_intern("exit!"), 0);
  }

  rb_funcall(rb_mProcess, rb_intern("waitall"), 0);

  printf("final srid: %i\n", GEOSGetSRID(geom)); // we always get `final srid: 1`
```

keithdoggett:
@BuonOmo yes I think that test works and makes sense.

This looks good. Feel free to merge whenever.

PR #331, opened 2022-09-26
refactor(geos_capi): use GEOS wkrep family by default

BuonOmo:
The option that was before only accessible by using the `:geos` symbol as a factory value for `wk[bt]_(generator|parser):` is now the default option. This makes wkrep way faster, and helps with our goal of having a fast geos implementation.

```ruby
require "benchmark/ips"
require_relative "lib/rgeo"

fast_factory = RGeo::Geos.factory
slow_factory = RGeo::Geos.factory(
  wkb_parser: {},
  wkt_parser: {},
  wkb_generator: {},
  wkt_generator: {},
)

data = IO.read(File.join(__dir__, "test", "wkrep", "data.wkt"))

Benchmark.ips do |x|
  x.report("geos") {fast_factory.parse_wkb(fast_factory.parse_wkt(data).as_binary).as_text}
  x.report("ruby") {slow_factory.parse_wkb(slow_factory.parse_wkt(data).as_binary).as_text}
  x.compare!
end
```

```
                geos   163.000  i/100ms
                ruby    22.000  i/100ms
                geos      1.622k (± 1.5%) i/s -      8.150k in   5.025429s
                ruby    217.877  (± 0.9%) i/s -      1.100k in   5.049089s

Comparison:
                geos:     1622.1 i/s
                ruby:      217.9 i/s - 7.45x  (± 0.00) slower
```

Signed-off-by: Ulysse Buonomo <buonomo.ulysse@gmail.com>


ext/geos_c_impl/factory.c
```diff
@@ -44,7 +44,7 @@ notice_handler(const char* fmt, ...)
 #endif
 
 static void
-error_handler(const char* fmt, ...)
+NORETURN(error_handler)(const char* fmt, ...)
```
keithdoggett, line 47: What's the purpose of this? Never seen it before so just curious.

keithdoggett (approved):
Left one question. And I have another one more generally here.

So when a `wk(t|b)_*` variable is `nil` in the factory creation method it will default to using the GEOS implemented version in the factory?

keithdoggett:
Looks like there's some minor differences between how the ruby parser and the GEOS parser handle floats

ext/geos_c_impl/factory.c
BuonOmo: It is a ruby implementation version of the C23 `noreturn` keyword (C17 ~ `_Noreturn` in some compilers) that tells the compilers that this function will never return.

This helps get rid of a warning (`warning: function 'error_handler' could be declared with attribute 'noreturn' [-Wmissing-noreturn]`) and very probably helps the compiler optimizing.

ext/geos_c_impl/factory.c
```diff
@@ -557,9 +559,9 @@ cmethod_factory_create(VALUE klass,
 
     if (context) {
       data->geos_context = context;
-      data->flags = NUM2INT(flags);
-      data->srid = NUM2INT(srid);
-      data->buffer_resolution = NUM2INT(buffer_resolution);
+      data->flags = RB_NUM2INT(flags);
```
BuonOmo, line 562: NUM2INT is the old naming, lets be consistent with the new naming!

ext/geos_c_impl/geometry.c
```diff
@@ -759,8 +759,8 @@ method_geometry_buffer_with_style(VALUE self,
                             self_geom,
                             rb_num2dbl(distance),
                             RGEO_FACTORY_DATA_PTR(factory)->buffer_resolution,
-                            rb_num2int(endCapStyle),
-                            rb_num2int(joinStyle),
+                            RB_NUM2INT(endCapStyle),
```
BuonOmo, line 762: The `rb_num2int` function actually returns a long, where the `RB_NUM2INT` macro try to return an int, and cast the long into int if failed

test/support/minitest/assert_wkt_similar.rb
```diff
@@ -0,0 +1,30 @@
...
+        end
+
+        def parser
+          @parser ||= RGeo::WKRep::WKTParser.new
+        end
+
+        def generator
+          @generator ||= RGeo::WKRep::WKTGenerator.new
+        end
+      end
```
keithdoggett, line 24: I guess the only issue with this is that it relies on another part of the package being implemented correctly to pass our tests. I think this is a smart way to handle it and is almost definitely fine since the `WKTParser` and `WKTGenerator` classes seldom change, but could potentially cause tests to seemingly randomly fail later if we were to introduce a bug in either of these.

BuonOmo: @keithdoggett I thought about that as well, my excuse is that this should not be buggy, and should also be tested anyway, so if tests are failing for that reason, we can be glad they fail and take this opportunity to consolidate.

That's just an excuse though, of course I'd rather have tests less intricated.. I'll look for how the wkt readers and writers are made in GEOS, and options we could use to have something more similar

BuonOmo:
@keithdoggett tests should pass now :)

test/support/minitest/assert_wkt_similar.rb
BuonOmo: There are definitely some options: https://libgeos.org/doxygen/geos__c_8h.html#a84d8fd4c78746d91fe0d6cffe3287ea8 (See GEOSWKTWriter_ methods for instance)

I don't know that we want to go much further down that road for now. Maybe we could find some kind of harmony between ruby and geos parser, and have the same interface for all ?

keithdoggett: Yeah I don't think it's necessary to do right now. This solution is likely going to suffice and assuming we have tested the `WKTParser` and `WKTGenerator` classes then it shouldn't be much of an issue because breaking changes to them would cause their tests to fail too.

PR #332, opened 2022-09-27
Move `std=c17` flag into optional flags

keithdoggett:
### Summary

The `std=c17` flag is only supported on gcc versions >= 9, which older distros of linux do not use by default. There's currently no features that require the use of c17 instead of c11, so we will relax this requirement and move that flag into the `MAINTAINER_MODE` group of `CFLAGS`.

if changes are made in the future that are not backwards compatible with c11, we will move it back out and note the gcc version requirement in the README.

### Other Information

Also adds `MAINTAINER_MODE` to the CI.


BuonOmo (commented):
One issue with using `MAINTAINER_MODE` in the CI: now tests are not ran in an environment that correspond to the real usage.

One solution I have in mind would be to set it for PRs to have early feedback, but not set it in the main branch, so when we push new versions of the gem we know they work in the final env. 

Or we could have yet another variable for CI, that doesn't set `-DDEBUG` or `-ggdb`, which both changes the build

BuonOmo (approved):

.github/workflows/CI.yml
```diff
@@ -61,6 +64,9 @@ jobs:
         run: |
           sudo apt-add-repository ppa:ubuntugis/ubuntugis-unstable
           sudo apt-get install -yqq libgeos-dev valgrind
+      - name: Set Maintainer Mode
+        if: ${{ github.event.pull_request }}
+        run: export MAINTAINER_MODE=1
```
BuonOmo, line 69: Why the need to set that in memcheck as well?

keithdoggett: It's not strictly necessary, but I figured we should add it here as well since it still compiles everything and runs the tests while also checking for leaks.

BuonOmo: So the issue I have with that is that memcheck is already slower to run. However, since tests on mac are even slower, and we are still around the 1mn mark, LGTM :)

keithdoggett: I think it's fine to remove it from that then. As long as we have it on the main test runner we'll be good.

PR #334, opened 2022-10-11
feat(geos): Add `Geometry#voronoi_diagram`

### LINKED ISSUE #93: Feature proposal: Voronoi tesselation
I'm currently using rgeo for a large Web based GIS project which needs to perform a Voronoi computation on a set of points and it ocurred to me that I could write it natively inside the rgeo gem.

I'm thinking of a simple interface which works on a geometry collection which includes only points and returns another geometry collection consisting of the polygons generated by the tesselation process. May be it could be a new method for the geometry collection class.

It would work on 2 dimensional geometries and would have the possibility of using euclidean and manhattan distance.

I'd love to add this new feature to the rgeo gem and would really appreciate any guidance, suggestions, comments and any other type of feedback.

I'm currently checking out the Fortune's algorithm.

Regards

BuonOmo:
Fixes #93

### DIFF
--- a/lib/rgeo/geos/capi_feature_classes.rb
+++ b/lib/rgeo/geos/capi_feature_classes.rb
@@ -75,7 +75,7 @@ module RGeo
       # The source may be any geometry type. All vertices of the geometry will be
       # used as the input points to the diagram.
       #
-      # @param enevelope [RGeo::Feature::Geometry | nil]
+      # @param envelope [RGeo::Feature::Geometry | nil]
       #   The +envelope+ keyword argument provides an envelope to use to clip the
       #   resulting diagram. If +nil+, it will be calculated automatically.
       #   The diagram will be clipped to the *larger* of the provided envelope


ext/geos_c_impl/geometry.c
```diff
@@ -1158,6 +1158,43 @@ method_geometry_polygonize(VALUE self)
...
+  Check_Type(tolerance, T_FLOAT);
+
+  result = Qnil;
+  self_data = RGEO_GEOMETRY_DATA_PTR(self);
+  self_geom = self_data->geom;
+
+  if (self_geom) {
+    if (RB_TEST(env)) {
+      Check_TypedStruct(env, &rgeo_geometry_type);
+      env_geom = rgeo_convert_to_geos_geometry(self_data->factory, env, Qnil);
```
BuonOmo, line 1183: This method could return NULL. In the spirit of v3, shouldn't we raise in stead?

lib/rgeo/geos/capi_feature_classes.rb
```diff
@@ -69,6 +69,44 @@ def as_text
...
+      #   The test +test_voronoi_diagram_tolerance+ in +test/common/voronoi_tests.rb+
+      #   shows more details.
+      #
+      # @param only_edges [true|false]
+      #   If the +only_edges+ keyword argument is +false+ a collection of +Polygon+s
+      #   will be returned. Otherwise a collection of +LineString+ edges is returned.
+      #
+      # @return [RGeo::Feature::Geometry]
+      #   if +only_edges+ is true, this will return a LineString or MultiLineString.
+      #   Otherwise, it will be a GeometryCollection of polygons.
```
BuonOmo, line 99: The fact that it has 3 different possible return values may be bothering, WDYT?

keithdoggett: I think that's fine. There's other methods like `union` that will return different types depending on the resultant geometry graph so this isn't unusual. Does accounting for the different types make it more difficult to handle in the method?

BuonOmo: Not at all. It may just make it more difficult in the usage. Especially the linestring vs multilinestring seems tricky. But to be fair, this may be a geos issue since it also affects geos and shapely!

ext/geos_c_impl/geometry.c
BuonOmo: It is not, but that's my point! I think this debate should be part of another PR anyway. But here it can affect the behaviour:

I run `geom.voronoi_diagram(envelope: poly)`, and I expect the envelope to be `poly`, but since the conversion silently failed, the envelope param is actually ignored.

I can solve the local issue by raising in case of null value, and solve the global problem in a later PR. Or just do nothing here and accept the potential silent failure here, which is anyway not that likely to happen (`Feature.cast` would need to return nil)

lib/rgeo/geos/capi_feature_classes.rb
BuonOmo: https://github.com/libgeos/geos/pull/702 Geos PR

ext/geos_c_impl/geometry.c
keithdoggett: > I can solve the local issue by raising in case of null value, and solve the global problem in a later PR

I think this is the best approach. I agree that this behavior should be the standard since it's more obvious that there was an issue. Of course it's still a similar amount of work for the user since they have to rescue now instead of checking for `nil` but I think being more explicit and actually raising instead of silently failing is better.

BuonOmo: I have added a local security for now, I'm thinking of quickly remove it in favor of the global solution. Otherwise this error could be present in various places in the code base

lib/rgeo/geos/capi_feature_classes.rb
```diff
@@ -101,7 +101,7 @@ def as_text
       # @see https://en.wikipedia.org/wiki/Voronoi_diagram
       # @see https://libgeos.org/doxygen/geos__c_8h.html#ace0b2fabc92d8457a295c385ea128aa5
       def voronoi_diagram(envelope: nil, tolerance: 0.0, only_edges: false)
-        Primary.voronoi_diagram(self, envelope, tolerance || 0.0, only_edges)
+        Primary.voronoi_diagram(self, envelope, Float(tolerance), only_edges)
```
BuonOmo, line 104: The change here allows one to pass any kind of numeric

ext/geos_c_impl/geometry.c
```diff
@@ -1158,6 +1158,46 @@ method_geometry_polygonize(VALUE self)
...
+        rb_raise(rb_eRGeoInvalidGeometry,
+                 "Could not use the given envelope with a GEOS factory. "
+                 "`RGeo::Feature.cast` failed.");
+    }
+    diagram = GEOSVoronoiDiagram(
+      self_geom, env_geom, rb_num2dbl(tolerance), RB_TEST(only_edges));
+
+    if (diagram == NULL) {
+      rb_raise(rb_eGeosError, "GEOS cannot create a voronoi diagram");
+    }
```
keithdoggett, line 1193: should we remove this and let the handler that we're going to implement in the wrappers take care of raising?

lib/rgeo/geos/capi_feature_classes.rb
```diff
@@ -69,6 +69,44 @@ def as_text
...
+      #
+      # @see https://en.wikipedia.org/wiki/Voronoi_diagram
+      # @see https://libgeos.org/doxygen/geos__c_8h.html#ace0b2fabc92d8457a295c385ea128aa5
+      def voronoi_diagram(envelope: nil, tolerance: 0.0, only_edges: false)
+        Primary.voronoi_diagram(self, envelope, Float(tolerance), only_edges)
+      rescue RGeo::Error::InvalidGeometry => e
+        message = "Could not create a voronoi_diagram with the specified inputs"
+        message += ". Try removing the `tolerance` parameter from ##{__method__}" if tolerance
+        raise e, message
+      end
```
keithdoggett, line 109: when would this get raised?

lib/rgeo/geos/ffi_feature_methods.rb
```diff
@@ -304,6 +304,22 @@ def point_on_surface
...
+          @fg_geom.voronoi_diagram(
+            envelope: fg,
+            tolerance: tolerance,
+            only_edges: only_edges
+          )
+        )
+      rescue ::Geos::GEOSException
+        message = "Could not create a voronoi_diagram with the specified inputs"
+        message += ". Try removing the `tolerance` parameter from ##{__method__}" if tolerance
+        raise RGeo::Error::InvalidGeometry, message
```
keithdoggett, line 320: should we raise an `InvalidGeometry` error or a `GeosError` here?

keithdoggett (approved):
Looks good. I just left a few questions. Primarily related to error handling and what types of errors we should be raising from Geos. I'd prefer to have errors from Geos raise `GeosError` instead of `InvalidGeometry` because with the validity changes it might be confusing to get `InvalidGeometry` errors on geometry creation when we've stated the intention is that will no longer happen.

lib/rgeo/geos/capi_feature_classes.rb
```diff
@@ -69,6 +69,44 @@ def as_text
         str
       end
       alias to_s as_text
+
+      # Constructs a Voronoi diagram from the vertices of the input geometry.
+      #
+      # The source may be any geometry type. All vertices of the geometry will be
+      # used as the input points to the diagram.
+      #
+      # @param enevelope [RGeo::Feature::Geometry | nil]
```
BuonOmo, line 78:
```suggestion
      # @param envelope [RGeo::Feature::Geometry | nil]
```

PR #340, opened 2022-10-24
V3 Docs

keithdoggett:
### Summary

* Add `NEWS.md` with discussion of the changes.
* Add Upgrading to RGeo Version 3.0 checklist.
* Update README

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,7 +14,7 @@ More details about usage and some technical details can be found in [doc/Geometr
 
 ### Coordinate System and Coordinate Transformations
 
-__Note: rgeo-proj4 version 4.0 will need to be used with RGeo 3.0__
+__Note: If you are using rgeo-proj4, you will need to upgrade it to version 4.0 to be used with RGeo 3.0__
 
 In an effort to improve the interface for transforming coordinates and decoupling the `rgeo-proj4` gem from the core RGeo library, we've changed how factories are given coordinate system information. The `proj4` option has been removed from all factory creation options and the `coord_sys` option will be relied on instead.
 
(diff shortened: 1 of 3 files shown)


BuonOmo (approved):
LGTM!

The validity part makes me think that we should finish handling all the cases of methods that return null rather than raising before having the final release.

Thinking for instance at https://github.com/rgeo/rgeo/blob/0fb78f64c55a936116d1d3c86a953dc03494c5e8/lib/rgeo/geos/ffi_factory.rb#L431 which really should not be a safe access, since `Feature.cast` should not return nil.

Same for the capi counter parts:

- https://github.com/rgeo/rgeo/blob/ec23154b99b1dd2b7900a3c2e5073c138c56acbe/ext/geos_c_impl/factory.c#L865-L867
- https://github.com/rgeo/rgeo/blob/ec23154b99b1dd2b7900a3c2e5073c138c56acbe/ext/geos_c_impl/factory.c#L834-L835

keithdoggett:
> The validity part makes me think that we should finish handling all the cases of methods that return null rather than raising before having the final release.

Yes, I agree. Do we just have to change these wrapper functions and instead of returning `NULL` raise a `GeosError`? Or do we have to do it on a method by method basis?

BuonOmo:
@keithdoggett I'd go for the wrapper. For the capi implementation, it is already raising if not null neither a geos geometry, hence we already have the correct implementation if we remove the null condition. For ffi, I'd say we can go the same way.

keithdoggett:
@BuonOmo sounds good. I can get a PR up for that soon.

PR #341, opened 2022-10-29
Raise Error After Failed Cast in GEOS Impls

keithdoggett:
### Summary

Raise errors instead of returning `nil` when objects cannot be properly cast to GEOS geometries in both the capi and FFI implementations.

These methods are primarily called when a method requires a second geometry (ex. `intersects?` takes a `rhs` argument). So now if invalid data is given in `rhs` an error will be raised.

### DIFF
--- a/ext/geos_c_impl/factory.c
+++ b/ext/geos_c_impl/factory.c
@@ -820,9 +820,10 @@ rgeo_wrap_geos_geometry_clone(VALUE factory,
 }
 
 const GEOSGeometry*
-rgeo_convert_to_geos_geometry(VALUE factory, VALUE obj, VALUE type)
+rgeo_convert_to_geos_geometry(VALUE factory, VALUE obj, VALUE type, int* state)
 {
   VALUE object;
+  VALUE msg;
 
   if (NIL_P(type) && RGEO_GEOMETRY_TYPEDDATA_P(obj) &&
       RGEO_GEOMETRY_DATA_PTR(obj)->factory == factory) {
@@ -831,11 +832,21 @@ rgeo_convert_to_geos_geometry(VALUE factory, VALUE obj, VALUE type)
     object =
       rb_funcall(rgeo_feature_module, rb_intern("cast"), 3, obj, factory, type);
   }
-  if (NIL_P(object))
-    rb_raise(rb_eRGeoInvalidGeometry,
-             "Unable to cast the geometry to the GEOS Factory");
+  if (NIL_P(object)) {
+    msg = rb_str_new_cstr("Unable to cast the geometry to the GEOS Factory");
+    rb_protect(
+      rb_exc_raise, rb_exc_new_str(rb_eRGeoInvalidGeometry, msg), state);
+  }
 
-  Check_TypedStruct(object, &rgeo_geometry_type);
+  if (*state) {
+    return NULL;
+  }
+
+  rb_protect(rgeo_check_geos_object, object, state);
+  if (*state) {
+    return NULL;
+  }
+
   return RGEO_GEOMETRY_DATA_PTR(object)->geom;
 }
 
@@ -850,6 +861,8 @@ rgeo_convert_to_detached_geos_geometry(VALUE obj,
   GEOSGeometry* geom;
   RGeo_GeometryData* object_data;
   const GEOSPreparedGeometry* prep;
+  VALUE msg;
+
   if (klasses) {
     *klasses = Qnil;
   }
@@ -863,12 +876,21 @@ rgeo_convert_to_detached_geos_geometry(VALUE obj,
                               type,
                               ID2SYM(rb_intern("force_new")),
                               ID2SYM(rb_intern("keep_subtype")));
-  if (*state || NIL_P(object)) {
-    rb_raise(rb_eRGeoInvalidGeometry,
-             "Unable to cast the geometry to the GEOS Factory");
+
+  if (NIL_P(object)) {
+    msg = rb_str_new_cstr("Unable to cast the geometry to the GEOS Factory");
+    rb_protect(
+      rb_exc_raise, rb_exc_new_str(rb_eRGeoInvalidGeometry, msg), state);
+  }
+  if (*state) {
+    return NULL;
   }
 
-  Check_TypedStruct(object, &rgeo_geometry_type);
+  rb_protect(rgeo_check_geos_object, object, state);
+  if (*state) {
+    return NULL;
+  }
+
   object_data = RGEO_GEOMETRY_DATA_PTR(object);
   geom = object_data->geom;
   if (klasses) {
--- a/ext/geos_c_impl/factory.c
+++ b/ext/geos_c_impl/factory.c
@@ -833,7 +833,7 @@ rgeo_convert_to_geos_geometry(VALUE factory, VALUE obj, VALUE type, int* state)
   }
   if (NIL_P(object)) {
     rb_protect(
-      rb_exc_raise,
+      rb_exc_raise_value,
       rb_exc_new_cstr(rb_eRGeoInvalidGeometry,
                       "Unable to cast the geometry to the GEOS Factory"),
       state);
@@ -843,7 +843,12 @@ rgeo_convert_to_geos_geometry(VALUE factory, VALUE obj, VALUE type, int* state)
     return NULL;
   }
 
-  rb_protect(rgeo_check_geos_object, object, state);
+  if (!rgeo_is_geos_object(object)) {
+    rb_protect(rb_exc_raise_value,
+               rb_exc_new_cstr(rb_eRGeoError, "Not a GEOS Geometry object."),
+               state);
+  }
+
   if (*state) {
     return NULL;
   }
@@ -879,7 +884,7 @@ rgeo_convert_to_detached_geos_geometry(VALUE obj,
 
   if (NIL_P(object)) {
     rb_protect(
-      rb_exc_raise,
+      rb_exc_raise_value,
       rb_exc_new_cstr(rb_eRGeoInvalidGeometry,
                       "Unable to cast the geometry to the GEOS Factory"),
       state);
@@ -888,7 +893,11 @@ rgeo_convert_to_detached_geos_geometry(VALUE obj,
     return NULL;
   }
 
-  rb_protect(rgeo_check_geos_object, object, state);
+  if (!rgeo_is_geos_object(object)) {
+    rb_protect(rb_exc_raise_value,
+               rb_exc_new_cstr(rb_eRGeoError, "Not a GEOS Geometry object."),
+               state);
+  }
   if (*state) {
     return NULL;
   }
--- a/ext/geos_c_impl/factory.c
+++ b/ext/geos_c_impl/factory.c
@@ -832,7 +832,8 @@ rgeo_convert_to_geos_geometry(VALUE factory, VALUE obj, VALUE type)
       rb_funcall(rgeo_feature_module, rb_intern("cast"), 3, obj, factory, type);
   }
   if (NIL_P(object))
-    return NULL;
+    rb_raise(rb_eRGeoInvalidGeometry,
+             "Unable to cast the geometry to the GEOS Factory");
 
   Check_TypedStruct(object, &rgeo_geometry_type);
   return RGEO_GEOMETRY_DATA_PTR(object)->geom;
@@ -863,9 +864,11 @@ rgeo_convert_to_detached_geos_geometry(VALUE obj,
                               ID2SYM(rb_intern("force_new")),
                               ID2SYM(rb_intern("keep_subtype")));
   if (*state || NIL_P(object)) {
-    return NULL;
+    rb_raise(rb_eRGeoInvalidGeometry,
+             "Unable to cast the geometry to the GEOS Factory");
   }
 
+  Check_TypedStruct(object, &rgeo_geometry_type);
   object_data = RGEO_GEOMETRY_DATA_PTR(object);
   geom = object_data->geom;
   if (klasses) {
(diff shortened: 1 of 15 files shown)


ext/geos_c_impl/factory.c
```diff
@@ -832,7 +832,8 @@ rgeo_convert_to_geos_geometry(VALUE factory, VALUE obj, VALUE type)
       rb_funcall(rgeo_feature_module, rb_intern("cast"), 3, obj, factory, type);
   }
   if (NIL_P(object))
-    return NULL;
+    rb_raise(rb_eRGeoInvalidGeometry,
+             "Unable to cast the geometry to the GEOS Factory");
```
keithdoggett, line 836: Some tests were failing because they check that an `InvalidGeometry` error is raised rather than any error, so I left the conditional in for now. We can remove it all together and allow `Check_TypedStruct` to raise, but we'll need to fix the tests.

ext/geos_c_impl/factory.c
```diff
@@ -863,9 +864,11 @@ rgeo_convert_to_detached_geos_geometry(VALUE obj,
                               ID2SYM(rb_intern("force_new")),
                               ID2SYM(rb_intern("keep_subtype")));
   if (*state || NIL_P(object)) {
-    return NULL;
+    rb_raise(rb_eRGeoInvalidGeometry,
+             "Unable to cast the geometry to the GEOS Factory");
   }
 
+  Check_TypedStruct(object, &rgeo_geometry_type);
```
keithdoggett, line 871: May not be necessary, but added it for safety.

ext/geos_c_impl/geometry.c
```diff
@@ -421,19 +421,17 @@ method_geometry_disjoint(VALUE self, VALUE rhs)
...
-    if (rhs_geom) {
 #ifdef RGEO_GEOS_SUPPORTS_PREPARED2
-      prep = rgeo_request_prepared_geometry(self_data);
-      if (prep)
-        val = GEOSPreparedDisjoint(prep, rhs_geom);
-      else
+    prep = rgeo_request_prepared_geometry(self_data);
+    if (prep)
+      val = GEOSPreparedDisjoint(prep, rhs_geom);
+    else
```
keithdoggett, line 428: I removed all of the checks since we should no longer have a `NULL` `rhs_geom`.

ext/geos_c_impl/geometry_collection.c
```diff
@@ -66,9 +66,7 @@ create_geometry_collection(VALUE module, int type, VALUE factory, VALUE array)
   for (i = 0; i < len; ++i) {
     geom = rgeo_convert_to_detached_geos_geometry(
       rb_ary_entry(array, i), factory, cast_type, &klass, &state);
-    if (state || !geom) {
-      break;
-    }
```
keithdoggett, line 71: I think this is ok to remove since `rgeo_convert_to_detached_geos_geometry` will raise on this condition.

ext/geos_c_impl/polygon.c
```diff
@@ -253,12 +253,6 @@ cmethod_create(VALUE module,
   linear_ring_type = rgeo_feature_linear_ring_module;
   exterior_geom = rgeo_convert_to_detached_geos_geometry(
     exterior, factory, linear_ring_type, NULL, &state);
-  if (state) {
-    rb_exc_raise(rb_errinfo());
-  }
-  if (!exterior_geom) {
-    return Qnil;
-  }
```
keithdoggett, line 261: Again, I think this is safe to remove because `rgeo_convert_to_detached_geos_geometry` checks for both of these, but I'm not quite sure.

lib/rgeo/geos/ffi_feature_methods.rb
```diff
@@ -158,11 +158,10 @@ def make_valid
       def equals?(rhs)
         return false unless rhs.is_a?(RGeo::Feature::Instance)
         fg = factory.convert_to_fg_geometry(rhs)
-        if !fg
-          false
+
```
keithdoggett, line 161: similar situation here. We can remove the `nil` checks

ext/geos_c_impl/geometry_collection.c
```diff
@@ -64,8 +84,23 @@ create_geometry_collection(VALUE module, int type, VALUE factory, VALUE array)
...
   for (i = 0; i < len; ++i) {
-    geom = rgeo_convert_to_detached_geos_geometry(
-      rb_ary_entry(array, i), factory, cast_type, &klass, &state);
+    args.obj = rb_ary_entry(array, i);
+    args.factory = factory;
+    args.type = cast_type;
+    args.klasses = &klass;
+    args.state = &state;
+
+    geom = rb_protect(rgeo_convert_to_detached_geos_geometry_wrapper,
```
keithdoggett, line 93: Since we're raising now, this function and the `coord_seq_from_array` function had memory leaks because we allocated data before calling `rgeo_convert_*` in each of them. This wraps those calls in `rb_protect` so that we can handle the error, free the existing data and then re-raise.

I could go with `rb_rescue`, but `rb_protect` feels a little nicer to handle all of the error logic within the original caller.

ext/geos_c_impl/factory.c
```diff
@@ -863,9 +864,11 @@ rgeo_convert_to_detached_geos_geometry(VALUE obj,
                               ID2SYM(rb_intern("force_new")),
                               ID2SYM(rb_intern("keep_subtype")));
   if (*state || NIL_P(object)) {
-    return NULL;
+    rb_raise(rb_eRGeoInvalidGeometry,
+             "Unable to cast the geometry to the GEOS Factory");
```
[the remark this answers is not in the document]
BuonOmo: For (1.), if we are scared of using `rb_set_errinfo` (because `rb_protect` actually seems to do much more), we could:

```c
  if (NIL_P(object))
    rb_protect(rb_exc_raise, rb_exc_new_str("Unable to cast the geometry to the GEOS Factory"), state);

  if (*state)
    return NULL;
```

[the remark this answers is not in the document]
keithdoggett: I like this solution. We'd have to do a `NULL` check but at least we can raise right then and halt execution instead of letting it propagate.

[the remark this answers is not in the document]
keithdoggett: I'll look at this soon and see if any other issues come up with it, but this seems like a fairly straightforward fix. Do you suggest doing this for both of the `geos_convert_*` methods or just the `detached` version?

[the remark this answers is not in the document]
keithdoggett: I guess both of them get used in a similar way that we see as an issue here so I can do both.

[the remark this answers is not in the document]
keithdoggett: @BuonOmo I'm going through this a little more now and I'm not sure we can meet the guarantee of the `rgeo_convert_*` methods not raising since we use `Check_TypedStruct` to validate the data types which may raise.

I can modify the `detached` version so that if it's `NULL` or `state` is returned, the user should handle the error, and the `state` variable doesn't just disappear, but I don't think it makes sense to pull the `Check_TypedStruct` calls out of these functions. What do you think?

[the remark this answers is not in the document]
keithdoggett: https://github.com/rgeo/rgeo/commit/6f241060f45d0477e23492794c48e789d0dd3190 This is the commit where we a added that check btw.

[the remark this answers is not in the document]
keithdoggett: @BuonOmo looked at this some more. It seems like I can replace these `Check_TypedStruct` calls with the `rgeo_is_geos_object` or `rgeo_check_geos_object` functions we defined and wrap it in `rb_protect`.

keithdoggett:
@BuonOmo my most recent changes propagate `state` through the internal `rgeo_convert_*` functions and the caller must determine how to handle an error. This makes cleanup a lot easier in cases where some data may have already been allocated.

ext/geos_c_impl/factory.c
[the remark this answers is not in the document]
BuonOmo: @keithdoggett it seems like there was a lot more to dig than what I expected. Maybe raising is the simpler option?

If you feel like you have the problem under control however, I fully agree with your choices.

My last concern is about performances, they may be a bit degraded when using protect. But tbh I don't know, and I'm not sure for now how to handle that.

ext/geos_c_impl/factory.c
```diff
@@ -831,10 +832,21 @@ rgeo_convert_to_geos_geometry(VALUE factory, VALUE obj, VALUE type)
     object =
       rb_funcall(rgeo_feature_module, rb_intern("cast"), 3, obj, factory, type);
   }
-  if (NIL_P(object))
+  if (NIL_P(object)) {
+    msg = rb_str_new_cstr("Unable to cast the geometry to the GEOS Factory");
+    rb_protect(
+      rb_exc_raise, rb_exc_new_str(rb_eRGeoInvalidGeometry, msg), state);
```
BuonOmo, line 838: IDK if you had an issue, but you could avoid the temp ruby variable

```suggestion
    rb_protect(
      rb_exc_raise, rb_exc_new_cstr(rb_eRGeoInvalidGeometry, "Unable to cast the geometry to the GEOS Factory"), state);
```

ext/geos_c_impl/factory.c
```diff
@@ -862,7 +876,18 @@ rgeo_convert_to_detached_geos_geometry(VALUE obj,
                               type,
                               ID2SYM(rb_intern("force_new")),
                               ID2SYM(rb_intern("keep_subtype")));
-  if (*state || NIL_P(object)) {
+
+  if (NIL_P(object)) {
+    msg = rb_str_new_cstr("Unable to cast the geometry to the GEOS Factory");
```
BuonOmo, line 881: same here of course!

ext/geos_c_impl/geometry_collection.c
```diff
@@ -66,9 +66,14 @@ create_geometry_collection(VALUE module, int type, VALUE factory, VALUE array)
...
     geom = rgeo_convert_to_detached_geos_geometry(
       rb_ary_entry(array, i), factory, cast_type, &klass, &state);
-    if (state || !geom) {
-      break;
+    if (state) {
+      for (j = 0; j < i; j++) {
+        GEOSGeom_destroy(geoms[j]);
+      }
+      FREE(geoms);
+      rb_jump_tag(state);
```
BuonOmo, line 74: nice! Maybe we could remove all of my `rb_exc_raise(rb_errinfo())` with that, seems like the way to go!

ext/geos_c_impl/geometry.c
```diff
@@ -420,20 +421,22 @@ method_geometry_disjoint(VALUE self, VALUE rhs)
...
-      if (val == 0) {
-        result = Qfalse;
-      } else if (val == 1) {
-        result = Qtrue;
-      }
+      val = GEOSDisjoint(self_geom, rhs_geom);
+    if (val == 0) {
+      result = Qfalse;
+    } else if (val == 1) {
+      result = Qtrue;
```
BuonOmo, line 439: what do you think of

```suggestion

      result = GEOSDisjoint(self_geom, rhs_geom) ? Qtrue : Qfalse
```

BuonOmo (approved):
Seems like we have something much cleaner now!

ext/geos_c_impl/factory.c
keithdoggett: Ahh completely missed that there's a `cstr` version of the function. I'll change to that. Tried using a cstr in the function I used and it would compile, but segfaulted once it got to that line.

ext/geos_c_impl/geometry_collection.c
keithdoggett: I didn't change the ones that were already in there, but I will change to that now.

ext/geos_c_impl/factory.c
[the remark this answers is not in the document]
keithdoggett: Yeah I can look into `rb_protect` and see if performance is going to be really degraded. I think `rb_protect` is intended for calling ruby methods from C code, but the interface is much nicer since you don't have to define a callback function with specific args like you do with `rb_rescue`.

I would say from a "coding flow" standpoint it feels good right now and the 2 `rgeo_convert_*` functions have a clear error handling pattern through `state`, but I can investigate `rb_protect`.

keithdoggett:
@BuonOmo thanks for the review! The only thing that is a minor concern (although it doesn't seem to be a big deal) is that `rb_protect` expects the given function to return a `VALUE`. The compiler gives a warning about this, but as far as I can tell it doesn't cause any crashes or other issues.

If we want to match the return type, we could wrap `rb_raise`/`rb_exc_raise` in a function like this

```c
VALUE rb_raise_value(VALUE error_type, char* cstr) {
  rb_raise(error_type, cstr);
  return Qnil;
}
```

BuonOmo:
I think I'd add the method to avoid having too much warnings. However, not such if that will really silent the error: you might have another error then saying that `rb_raise_value` is not returning and should be declared as such

keithdoggett:
@BuonOmo I added a function that returns a `VALUE`. Similar to this use case from the ruby library

https://github.com/ruby/ruby/blob/d92f09a5eea009fa28cd046e9d0eb698e3d94c5c/ext/-test-/exception/ensured.c#L21-L33

I also cleaned up a few other warnings.

ext/geos_c_impl/ruby_more.h
```diff
@@ -12,6 +12,9 @@ RGEO_BEGIN_C
 VALUE
 rb_protect_funcall(VALUE recv, ID mid, int* state, int n, ...);
 
+VALUE
+rb_exc_raise_value(VALUE exc);
```
BuonOmo, line 16: Maybe add a short comment here for the reason?

BuonOmo (approved):

PR #342, opened 2022-10-31
Add `transform` method to `Geometry`

keithdoggett:
### Summary

Adds a method `transform` to `Geometry` that accepts the target factory as its sole argument. This is just a simple convenience method that wraps around `RGeo::Feature.cast`.

```ruby
require 'rgeo'
require 'rgeo/proj4'

fac1 = RGeo::Geos.factory(srid: 4326)
fac2 = RGeo::Geos.factory(srid: 3857)

pt1 = fac1.point(1, 1)

# do this
pt2 = pt1.transform(fac2)
# instead of
# pt2 = RGeo::Feature.cast(pt1, factory: fac2, project: true)
```

### Other Information

We discussed potentially accepting different types as the argument. For example an integer representing the SRID you want to transform to or a string defining the coordinate system.

The issue with this is that since all geometries need a factory, we would have to create a new factory each time `transform` is called, so we'd end up with N identical factories for N geometries. We might revisit the idea later on.


lib/rgeo/feature/geometry.rb
```diff
@@ -711,6 +711,18 @@ def +(rhs)
...
+      # Convenience method to transform/project a geometry
+      # to a different coordinate system from the geometry itself
+      # instead of the cast method.
+      #
+      # @note: Not an OGC SFS method
+      #
+      # @param [RGeo::Feature::Factory] other_factory
+      # @return [RGeo::Feature::Geometry]
+      def transform(other_factory)
+        Feature.cast(self, factory: other_factory, project: true)
```
BuonOmo, line 724: Why do we need to set `project: true` here ?

BuonOmo (approved):
This one is really neat! Just one small question (already asked)

lib/rgeo/feature/geometry.rb
keithdoggett: If we don't set `project: true`, it will not try to transform the coordinates and just re-wrap the object from its factory to the given factory.

PR #347, opened 2022-12-09
refactor(tests): Use modules to skip GEOS

BuonOmo:
This helps changing the behavior quickly, and clarify intents of `setup` methods.


keithdoggett (approved):
LGTM

Issue #336, opened 2022-10-11
Tackle rubocop todo list


BuonOmo:
The `.rubocop_todo.yml` file needs to be taken care of. Any PR in that regard is welcome!

thestelz:
So I was going to take a stab at this, but I either have an issue with my environment or something is wrong with the gemspec.

I ran `rubocop` and got:

```
Error: unrecognized cop or department Layout/LineContinuationLeadingSpace found in .rubocop.yml
Did you mean `Layout/MultilineOperationIndentation`?
```

I then wanted to confirm the right version of `rubocop` that was installed:

```
rubocop -v
1.28.2
```

That is when I realized that in the gemspec, it has `spec.add_development_dependency "rubocop", "~> 1.36.0"`, so I ran a `bundle install`. Then got:

```
bundle install
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Bundler could not find compatible versions for gem "ruby":
  In Gemfile:
    ruby

    rgeo was resolved to 3.0.0.pre.rc.3, which depends on
      ruby (>= 2.5.0)

    rubocop (~> 1.36.0) was resolved to 1.36.0, which depends on
      ruby (>= 2.6.0)
```

In the gemspec it has `spec.required_ruby_version = ">= 2.5.0"`, so should that be bumped up or am I doing something incorrectly?

On a side note, it doesn't appear I can install `2.6.0` either, but that might be an issue with my system.

```
rbenv install 2.6.0
Downloading ruby-2.6.0.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.6/ruby-2.6.0.tar.bz2
Installing ruby-2.6.0...

BUILD FAILED (macOS 12.6 using ruby-build 20211227)

Inspect or clean up the working tree at /var/folders/s7/ghx09h6s7j7flwp1hvjmqsp00000gn/T/ruby-build.20221014122627.85939.xRPfVh
Results logged to /var/folders/s7/ghx09h6s7j7flwp1hvjmqsp00000gn/T/ruby-build.20221014122627.85939.log

Last 10 log lines:
transform_mjit_header: SKIPPED to transform __inline_isfinitel
transform_mjit_header: SKIPPED to transform __inline_isfinited
transform_mjit_header: SKIPPED to transform __inline_isfinitef
transform_mjit_header: SKIPPED to transform __darwin_fd_clr
transform_mjit_header: SKIPPED to transform __darwin_fd_set
[...]

PR #351, opened 2023-01-25
Remove Naming/VariableNumber rubocop offense

haroon26:
### Summary

Fix `Naming/VariableNumber` offense
Updates Issue #336


keithdoggett:
Thanks @haroon26 I'm going to merge, but after that I'll edit the `History` file to break it out into a new `Current` section since we deployed version 3 already.

haroon26:
@keithdoggett Actually I have checked the log and found that author of commits was not correct. I have ammed my commits, but that eventually deleted your merging with `main`. Can you please check this?

PR #353, opened 2023-01-28
drop support for ruby 2.6 and test jruby

seuros:
### Summary

We are currently testing support for Ruby 2.6 to ensure compatibility with JRuby. However, JRuby now supports versions 2.7 to 3.1. I think upgrading the support to at least Ruby 2.7, which is nearing its end-of-life. Or even go directly to 3.0.

@keithdoggett , I want to give some maintenance shot to the repos. 
Can you enable ci for my pr ?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,11 +13,11 @@ jobs:
       matrix:
         ruby:
           - head
+          - '3.2'
           - '3.1'
           - '3.0'
           - '2.7'
-          - '2.6'
-          - "jruby-9.3.7.0"
+          - "jruby"
         os:
           - ubuntu
           # Windows users, feel free to open a PR :)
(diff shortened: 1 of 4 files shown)


keithdoggett:
Hey @seuros thank you for doing this and appreciate your help with the projects!

keithdoggett (approved):

keithdoggett:
@seuros this issue with `jruby` popped up on other RGeo gem's CIs. It seems like if we use `jruby` head there's some bug in an implementation and a loop doesn't work as expected. I didn't look into it too much but the failing CIs show where the bug is occuring.

I'm not too familiar with JRuby, but I might have some time to look into this issue in a few days.

seuros:
@keithdoggett i can look into it.

keithdoggett:
@seuros thanks! I got busy this week but should have some time soon if you can't get around to it either.

keithdoggett:
Hey @seuros checking back in on this. It appears there was a bug in jruby-9.4.0.0 that was causing the issue. I just tried locally with the latest version (9.4.2.0) and it works. Can you push or re-run the tests so we can ensure that it is working on the CI as well then we can merge everything.

Thanks!

seuros:
Yes

seuros:
I have few more pr to open

keithdoggett:
Awesome looking forward to it. If you need help with any of them let me know

Issue #350, opened 2023-01-25
Add Supported Geos Versions to CI


keithdoggett:
Currently we only test against the newest release of Geos from the `ubuntugis/ubuntugis-unstable` repository. Since we support more versions of Geos than that, we should add those versions to the CI matrix so we have a better understanding of how our C/FFI implementations work across all versions.

seuros:
Fixed in #392 

3.12 and 3.14.

PR #355, opened 2023-04-02
Test against a specific geos version

mylesboone:
### Summary

This implements #350 

### Other Information

More versions can be added. I'm not sure which versions are technically supported


mylesboone:
@keithdoggett Looks like macos and geos version 3.10+ work as expected.
geos version <3.10 fail tests everywhere. geos with ubuntu fails (and capi and ffi support not available)
Any ideas?

keithdoggett:
@mylesboone thank you for the work on this!

* Geos < 3.10 failing. In version 3.10 they redid some of the validity checks, so the tests that are failing expect one kind of error, but we're getting another. "Duplicate-Ring" vs. "Self-Intersection". I'm not sure the best way to handle this. We could check for either message or make a conditional based on the geos version. @BuonOmo what are your thoughts?
* Not sure about all the ubuntu builds failing. Quickly looked through the build logs and didn't notice anything failing during setup. I'll take a longer look through soon and see if I can figure it out.

BuonOmo:
@keithdoggett I'd say we follow Geos version, and adapt the test to the version with a conditional. I feel like having these kind of conditions directly in the codebase may lead to some weird behaviors later..

keithdoggett:
@BuonOmo I agree that's best too. @mylesboone So I guess what we can do is once we get ubuntu figured out, I'll open a new branch and this will be failing but we can merge into that and in that new branch we can add conditionals to the tests based on geos version.

BuonOmo:
Replaced by https://github.com/rgeo/rgeo/pull/362

PR #357, opened 2023-05-24
Update Github pages links to use github.io and use HTTPS

joshmcarthur:
### Summary

Github Pages has deprecated using github.com for pages since April 2021, and github.io should be used instead. This commit also updates protocols used in link URLs to use HTTPS rather than HTTP. I'm not sure how many people aim to find out more about RGeo via the github pages site, but since it showed up via search for me I thought it was worth fixing up the link to the docs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -60,13 +60,13 @@
 </ul><p>For more information:</p>
 
 <ul>
-<li>See the <a href="http://rgeo.github.com/rgeo/rdoc">documentation</a>
+<li>See the <a href="https://rgeo.github.io/rgeo/rdoc">documentation</a>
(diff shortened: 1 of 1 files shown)


BuonOmo:
Hey,

Thank you for contributing! I think we should just delete this branch as we have a dedicated repo for the website (https://github.com/rgeo/rgeo.github.io). Feel free to contribute on that repo, even more if you are a designer 😅 

Here's the new logo: https://www.reddit.com/r/ruby/comments/ryavb1/a_logo_for_rgeo/.

@keithdoggett ok for branch deletion ?

PR #361, opened 2023-08-23
fix(capi): Compatibility with GEOS 3.12

### LINKED ISSUE #360: With geos 3.12, the CAPI factory enables Z geometries
### Steps to reproduce

Install rgeo built against geos 3.12. Here is a Dockerfile to reproduce it.

```dockerfile
FROM ruby:3.2.2-bullseye

RUN apt update && apt install --yes cmake ruby-dev

RUN wget https://download.osgeo.org/geos/geos-3.12.0.tar.bz2
RUN tar -xf geos-3.12.0.tar.bz2
RUN mkdir geos-3.12.0/build && cd geos-3.12.0/build && cmake -DCMAKE_BUILD_TYPE=Release ..
RUN cd geos-3.12.0/build && make && make install
RUN ldconfig
RUN gem install rgeo

CMD ruby -e "require 'rgeo'; puts RGeo::Geos.capi_supported? ? RGeo::Cartesian.preferred_factory(srid: 4326).point(1, 1).as_text : 'CAPI not supported'"
```

In a shell:

```sh
docker build .  --tag 'rgeo_geos_3_12'
docker run rgeo_geos_3_12
```

### Expected behavior

Creates a geometry without Z. It displays `POINT (1 1)`

### Actual behavior

Creates a geometry with Z: `POINT Z (1 1 0)`

### System configuration

**Ruby version**:

3.2.2

**OS**:

Ubuntu bullseye in docker.
Geos 3.12 build from sources.

It also happens on MacOS with geos installed through brew.

BuonOmo:
The 3.12 include changes in the API responses.
Mainly, WBX Writers now default to three
dimensions instead of two. We follow the
changes, making no breaking change in the
API, but some changes in the responses
given by `#ccw?` and `#simplify_preserve_topology`.

See changes in libgeos:

- https://github.com/libgeos/geos/pull/784
- https://github.com/libgeos/geos/pull/878
- https://github.com/libgeos/geos/pull/908
- https://github.com/libgeos/geos/pull/915

Fixes #360


test/geos_capi/polygon_test.rb
```diff
@@ -90,7 +90,12 @@ def test_simplify_preserve_topology
     simplified = poly.simplify_preserve_topology(1)
     interior_points = simplified.interior_rings[0].points
 
-    assert_equal 5, interior_points.length
+    # https://github.com/libgeos/geos/pull/784
+    if Gem::Version.new(RGeo::Geos.version) >= Gem::Version.new("3.12.0")
```
keithdoggett, line 94: Maybe we add this check as a method in test_helper.rb in case we need to do this check in the future.

```ruby
def geos_version_gteq(version)
  Gem::Version.new(RGeo::Geos.version) >= Gem::Version.new(version)
end
```

keithdoggett (approved):
LGTM thanks for the quick fixes on this @BuonOmo. It's nice that we're able to keep all the versioning logic in the tests for this.

Related to this I should have some time to keep looking at the Geos gh test matrix soon as well.

BuonOmo:
I've applied the change, may I leave you the duty of merging if it suits you ?

PR #362, opened 2023-08-24
CI: Test Against Geos Versions

keithdoggett:
### Summary

Picking up from where we left off on #355.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,30 +21,53 @@ jobs:
         os:
           - ubuntu
           # Windows users, feel free to open a PR :)
+          # See issue #242, windows support would be welcome.
           # - windows
           - macos
-        exclude:
-          - os: windows # See issue #242, windows support would be welcome.
+        geos:
+          - '3.12.0'
+          - '3.11.2'
+          - '3.10.5'
+          - '3.9.4'
+          - '3.8.3'
     runs-on: ${{ matrix.os }}-latest
     continue-on-error: ${{ matrix.ruby == 'head' || matrix.os == 'windows' || matrix.os == 'macos' }}
-    name: Ruby ${{ matrix.ruby }} (${{ matrix.os }})
+    name: Ruby ${{ matrix.ruby }}, Geos ${{ matrix.geos }} (${{ matrix.os }})
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/cache@v2
+        with:
+          id: geos-cache
+          path: ./geos-${{ matrix.geos }}
+          key: geos-${{ matrix.geos }}-${{ matrix.os }}
+      - name: Install Packages (Linux)
+        if: matrix.os == 'ubuntu'
+        run: |
+          sudo apt-add-repository ppa:ubuntugis/ubuntugis-unstable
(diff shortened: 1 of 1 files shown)


keithdoggett:
@BuonOmo I was able to get this to install and work across all versions on ubuntu systems but I'm having issues with cmake not working on some of the macos runners. It's strange because the issue is only happening on some of the runners. I can look into this more in a bit, but if you have any ideas/suggestions let me know.

BuonOmo:
@keithdoggett by looking at the logs I can see that installing cmake installs version 3.27.2, but then while installing Geos we're looking for the version 3.27.3.

Idk what software is asking for a specific version, but that is for sure the issue.

.github/workflows/CI.yml
```diff
@@ -27,30 +27,62 @@ jobs:
...
-          ruby-version: ${{ matrix.ruby }}
-          bundler-cache: true
-      - name: Install Geos (Linux)
+      - name: Install Packages (Linux)
         if: matrix.os == 'ubuntu'
         run: |
           sudo apt-add-repository ppa:ubuntugis/ubuntugis-unstable
-          sudo apt-get install -yqq libgeos-dev
-      - name: Install Geos (Mac)
+          sudo apt-get install curl libcurl4-openssl-dev libssl-dev cmake
```
BuonOmo, line 48: I'd keep the `-yqq` because this avoids having logs too large, which we don't need anyway :)

keithdoggett: done

keithdoggett:
@BuonOmo it's almost working now. For whatever reason the macos jruby builds can't find the libgeos_c library when loading it via FFI. I even tried specifying the path explicitly with an environment variable, but no luck. Not sure if there's some JRuby quirk I'm just not aware of that's causing the issue. Can investigate more in a little while.

.github/workflows/CI.yml
```diff
@@ -27,30 +27,62 @@ jobs:
...
+          cd geos-${{ matrix.geos }}/build
+          sudo cmake --build . --target install
+          cd ../../
+      - name: ldconfig
+        if: matrix.os == 'ubuntu'
+        run: sudo ldconfig
+      - name: Set GEOS_LIBRARY_PATH (macos jruby)
+        if: matrix.os == 'macos' && matrix.ruby == 'jruby'
+        run: export GEOS_LIBRARY_PATH=/usr/local/lib/libgeos_c.dylib
+      - uses: ruby/setup-ruby@v1
```
BuonOmo, line 82: I tried on my mac, and I didn't need this `GEOS_LIBRARY_PATH` variable in my path

keithdoggett: yeah I'll try again I'm pretty stumped as to why this it isn't recognizing Geos. I shouldn't have to specify a path but clearly something's not configured correctly.

BuonOmo:
@keithdoggett thanks for getting that ball rolling !

It feels to me like a lot of redundancy both in execution (some tests are played twice) and in code. I don't have access to a laptop here to dig a bit the subject. Although I'm sure we could find a way to make this DRYer and still quite consistent and clear. What do you think?

PR #364, opened 2023-10-11
feat: Segmentize for GeosCapi geometries.

arkirchner:
### Summary
Modify geometry to have no segment longer then the max segment distance.

Feature is equal to:
[ST_Segmentize](https://postgis.net/docs/ST_Segmentize.html)
[shapely segmentize](https://shapely.readthedocs.io/en/latest/reference/shapely.GeometryCollection.html#shapely.GeometryCollection.segmentize)

### Other Information
- ffi support is planned: https://github.com/dark-panda/ffi-geos/pull/34/files
- I decided against calling the method like the Geos feature because it is popular as segmentize through PostGIS and shapely.

### DIFF
--- a/History.md
+++ b/History.md
@@ -6,6 +6,7 @@
 * Fix Naming/VariableNumber from rubocop_todos. (@haroon26) #351
 * Fix Lint/MissingSuper from rubocop_todos. (@haroon26) #352
 * Drop support for Ruby 2.6. (@seuros) #353
+* Add `segmentize` method to CAPI geometires (@arkirchner) #364
 
 **Bug Fixes**
 
--- a/README.md
+++ b/README.md
@@ -43,8 +43,7 @@ RGeo works with the following Ruby implementations:
 
 Some features also require the following:
 
-*   GEOS 3.2 or later is highly recommended. (3.3.3 or later preferred.) Some
-    functions will not be available without it. This C/C++ library may be
+*   GEOS 3.10 or later is highly recommended. Some functions will not be available without it. This C/C++ library may be
     available via your operating system's package manager (`sudo aptitude
     install libgeos-dev` for debian based Linux distributions, `yum install geos geos-devel` for redhat based Linux distributions), or you can
     download it from http://trac.osgeo.org/geos
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ RGeo works with the following Ruby implementations:
 
 Some features also require the following:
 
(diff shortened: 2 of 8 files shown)


keithdoggett (commented):
@arkirchner thanks for the PR! This looks good, but I have 2 notes

1. Looks like there's some formatting issues. You can use `bin/clang-format format` to fix it.
2. Do you know what version of Geos this function is available in? We may want to add guards around the method. 

@BuonOmo had done some work around getting C methods added in a better way so maybe he has some thoughts too

BuonOmo (commented):
I think we should raise if the given maximum length isn't a number. Also I don't know what happens when we give 0 or a negative value to the segment. I know that the Geos doc is not talking about any errors. But there still might be one!

Also I didn't understand your last point in the PR description (cannot quote you from my phone)

arkirchner:
@keithdoggett 

> Looks like there's some formatting issues. You can use bin/clang-format format to fix it.
I was able to fix the format.

> Do you know what version of Geos this function is available in? We may want to add guards around the method.
It is available from **GEOS 3.10** onwards. I updated the README to mention this. How can I add a guard.

arkirchner:
@BuonOmo
I added some tests for invalid inputs. The existing error messages are quite good. 🤔

keithdoggett:
@BuonOmo I was referencing the work you had done on this PR https://github.com/rgeo/rgeo/pull/334. I know part of the goal of that was to better understand how to add more Geos functions in a standardized manner, so I thought you might have some insights here.

arkirchner:
@keithdoggett Thank you for the detailed explanation. I wrapped both the method definition and the function to not cause any issues when compiling.

I reverted the change to the README because Geos version requirements do not change when using this method.

How about the tests. Should they be skipped for unsupported Geos versions?

keithdoggett:
@arkirchner thank you for implementing that! And yes if we could skip the test based on version too that would be great.

arkirchner:
@keithdoggett If you have time could you look at my test solution?

keithdoggett:
@arkirchner LGTM once that RuboCop error is fixed I'll merge! Thanks for the work on this and apologies about the delays in review.

arkirchner:
@keithdoggett 
Thank you very much. I fixed Rubocop and updated the History.

Should I squash everything once before merging?

keithdoggett (approved):

keithdoggett:
LGTM sorry about the delays. Thanks @arkirchner

PR #366, opened 2023-11-24
feat: add simplify_polygon_hull method

oleksii-leonov:
### Summary

The feature is equal to [ST_SimplifyPolygonHull](https://postgis.net/docs/ST_SimplifyPolygonHull.html) in PostGIS.

> Computes a simplified topology-preserving outer or inner hull of a polygonal geometry.
> An outer hull completely covers the input geometry.
> An inner hull is completely covered by the input geometry.
> The result is a polygonal geometry formed by a subset of the input vertices.
> MultiPolygons and holes are handled and produce a result with the same structure as the input.
> https://postgis.net/docs/ST_SimplifyPolygonHull.html

Utilizes the `GEOSPolygonHullSimplify` method introduced in [GEOS 3.11.0](https://github.com/libgeos/geos/releases/tag/3.11.0).

### Other Information

- https://github.com/libgeos/geos/issues/603
- https://github.com/libgeos/geos/pull/622
- https://github.com/locationtech/jts/pull/861
- https://github.com/libgeos/geos/commit/1b3521ccfb4de7fb4bd15ebfa4772b2da8155f30

### DIFF
--- a/ext/geos_c_impl/geometry.c
+++ b/ext/geos_c_impl/geometry.c
@@ -831,34 +831,6 @@ method_geometry_simplify_preserve_topology(VALUE self, VALUE tolerance)
   return result;
 }
 
-#ifdef RGEO_GEOS_SUPPORTS_POLYGON_HULL_SIMPLIFY
-static VALUE
-method_geometry_simplify_polygon_hull(VALUE self,
-                                      VALUE vertex_fraction,
-                                      VALUE is_outer)
-{
-  VALUE result;
-  RGeo_GeometryData* self_data;
-  const GEOSGeometry* self_geom;
-  VALUE factory;
-
-  unsigned int is_outer_uint = RTEST(is_outer) ? 1 : 0;
-
-  result = Qnil;
-  self_data = RGEO_GEOMETRY_DATA_PTR(self);
-  self_geom = self_data->geom;
-  if (self_geom) {
-    factory = self_data->factory;
-    result = rgeo_wrap_geos_geometry(
-      factory,
-      GEOSPolygonHullSimplify(
-        self_geom, is_outer_uint, rb_num2dbl(vertex_fraction)),
-      Qnil);
-  }
-  return result;
-}
-#endif
-
 static VALUE
 method_geometry_convex_hull(VALUE self)
 {
@@ -1357,14 +1329,6 @@ rgeo_init_geos_geometry()
     geos_geometry_methods, "make_valid", method_geometry_make_valid, 0);
   rb_define_method(
     geos_geometry_methods, "polygonize", method_geometry_polygonize, 0);
-
-#ifdef RGEO_GEOS_SUPPORTS_POLYGON_HULL_SIMPLIFY
-  rb_define_method(geos_geometry_methods,
-                   "simplify_polygon_hull",
-                   method_geometry_simplify_polygon_hull,
-                   2);
-#endif
-
 #ifdef RGEO_GEOS_SUPPORTS_DENSIFY
   rb_define_method(
     geos_geometry_methods, "segmentize", method_geometry_segmentize, 1);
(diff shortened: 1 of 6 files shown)


ext/geos_c_impl/geometry.c
```diff
@@ -831,6 +831,32 @@ method_geometry_simplify_preserve_topology(VALUE self, VALUE tolerance)
...
+    factory = self_data->factory;
+    result = rgeo_wrap_geos_geometry(
+      factory,
+      GEOSPolygonHullSimplify(
+        self_geom, is_outer_uint, rb_num2dbl(vertex_fraction)),
+      Qnil);
+  }
+  return result;
+}
+
```
keithdoggett, line 859: Should we add method guards around this too? Not sure if having the undefined `GEOSPolygonHullSimplify` symbol will break compilation in older versions of not guarded or would only cause a segfault if it's called.

oleksii-leonov: @keithdoggett , you are right 👍 
It's safer to wrap the method with a guard.

I have updated the PR.

keithdoggett:
Thanks @oleksii-leonov I'll give @BuonOmo another day or 2 to review, but if he's not available I'll merge it in.

BuonOmo (commented):
Thanks for the beautiful tests ! And for the overall feature. As always my concern is about the geos API. Can the hull generation method fail? And if it does, what happens?

Unfortunatily I cannot deep dive into geos codebase with my smartphone. So have verified that @oleksii-leonov ? You can also take advantage of running tests with `MAINTAINER_MODE` for more insights.

If you can't be sure, maybe @keithdoggett you could have a second pair of eyes on this ? 👀

Besides that, LGTM

oleksii-leonov:
@BuonOmo 

It looks like the PolygonHullSimplifier (https://github.com/libgeos/geos/blob/3.12/src/simplify/PolygonHullSimplifier.cpp) has only one possible exception:
`throw util::IllegalArgumentException("Input geometry must be polygonal")`:
https://github.com/libgeos/geos/blob/5b60b76fc0dbab04ed2488f78c4239a8b82657b8/src/simplify/PolygonHullSimplifier.cpp#L105C1-L105C1

It's handled correctly by RGeo:
```ruby
RGeo::Geos.factory.parse_wkt("LINESTRING (0 0, 0 6, 6 6, 6 0, 0 0)").simplify_polygon_hull(0.1, true)
# (irb):4:in `simplify_polygon_hull': Input geometry must be polygonal (RGeo::Error::InvalidGeometry)
```

I could move `simplify_polygon_hull` from `ext/geos_c_impl/geometry.c` to `ext/geos_c_impl/polygon.c`.

BuonOmo:
@oleksii-leonov thank you for verifying. I would add a test to reflect this.

On moving the method I think you are right, it would indeed be a better place.

keithdoggett (approved):
Thanks for moving the method @oleksii-leonov and checking for the exception. I think this is good to merge now. WDYT @BuonOmo

BuonOmo (commented):
@oleksii-leonov I’m sorry I don’t see the test catching the error we’ve talked about, have I missed it? I think it is an important inclusion!

oleksii-leonov:
@BuonOmo, `Input geometry must be polygonal` exception could happen only if you pass non-polygonal geometry to PolygonHullSimplifier.

Since I have moved `simplify_polygon_hull` from `ext/geos_c_impl/geometry.c` to `ext/geos_c_impl/polygon.c`, it's impossible to trigger that exception.

BuonOmo:
@oleksii-leonov indeed got you! Sorry for the delay, and thank you for the clear contribution :taco:

oleksii-leonov:
@BuonOmo , no worries :)
I have rebased the PR.

BuonOmo (approved):

PR #370, opened 2024-07-25
allow to pass kwarg params to make_valid method.

balbesina:
### Summary

Default makeValid provides very sad (and useless) resolution for intersected polygons of multipolygon. However, i dont wanna make sql requests with `SELECT ST_MakeValid(geom, 'method=structure')` from app to db. 

Here is an extension of `#make_valid` that accepts kwargs. "Strange" arguments choice mimics the behaviour of [SQL func](https://postgis.net/docs/ST_MakeValid.html)

[symbol] `method`=:linework|:structure
applies method structure only if :structure was provided. else acts as if :linework was given.

[bool] `keep_collapsed`=true|false
applies keepCollapsed when set only for method structure. else acts as if false was given.

### Other Information

It is still possible to not pass anything, persisting existing functionality.

I'm not sure what made devs working on SQL func to use such a complex way to define method with a string variable. Mb we should use bool flag (e.g. `method_structure`) instead? Or it's planned to extend makeValid with more possible methods? RN it works like linework=0 by default and can also take structure=1. So it's kinda bool. In [geos_c.h](https://github.com/libgeos/geos/blob/6201229a66ae3eef2b0c62cc39dbab48b2e3df7e/capi/geos_c.h.in#L1475) it is an enum with two variants. keepCollapsed on the other hand is simple int 1/0 - typical flag.


keithdoggett (changes requested):
@balbesina thanks for the PR! I think the code looks good, but can you add some tests to demonstrate/confirm usage?

keithdoggett:
@BuonOmo the arg parsing looks good to me but I know you're more familiar with this than I am so maybe you have a different opinion

BuonOmo:
@keithdoggett I think it is fine. I'll review further to make sure we are not generating some memory leak in the process once you add tests @balbesina, thank you for the contribution!

BuonOmo:
@balbesina you still have the lint test failing and I believe the memcheck will fail as well since you didn't change the code. Do you need any help with these? I suggest running these tests on your forks so you don't have to wait for approval every time

balbesina:
@BuonOmo thx for tips. i've reworked code a little bit to make it cleaner. also thx to geos guys answers i applied default params for non-param make_valid call (should be method=linework, keepCollapsed=**true**).

i ran clang-format and the change test is blaming me for (removing spaces before comment in file factory.c:1013) was done by `bin/clang-format`. if you want - i can roll it back.

for memcheck i think it should be an exclusion added to valgrind/suppressions, because i do call `GEOSMakeValidParams_destroy` - exactly as it is used in geos tests: [link1](https://github.com/libgeos/geos/blob/85b06abbace70f47c699b081be951930edfd1d50/tests/unit/capi/GEOSMakeValidTest.cpp#L75), [link2](https://github.com/libgeos/geos/blob/85b06abbace70f47c699b081be951930edfd1d50/tests/unit/capi/GEOSMakeValidTest.cpp#L91)

but when i try to build docker image for valgrind it fails in test/valgrind/Dockerfile:14 `COPY . .`
ERROR: failed to solve: cannot copy to non-directory: /var/lib/docker/overlay2/long-random-hash/merged/bin

any advice on that?

BuonOmo:
> i ran clang-format and the change test is blaming me for (removing spaces before comment in file factory.c:1013) was done by bin/clang-format

Please do, I'll check that later. Thanks for reporting.

> for memcheck i think it should be an exclusion added to valgrind/suppressions, because i do call GEOSMakeValidParams_destroy - exactly as it is used in geos tests: [link1](https://github.com/libgeos/geos/blob/85b06abbace70f47c699b081be951930edfd1d50/tests/unit/capi/GEOSMakeValidTest.cpp#L75), [link2](https://github.com/libgeos/geos/blob/85b06abbace70f47c699b081be951930edfd1d50/tests/unit/capi/GEOSMakeValidTest.cpp#L91)

The error can come from some other place. I'll check that. I also see that you have a failure in ruby 2.7 but maybe can just remove support for that branch as it is not maintained anymore. I'll push that upstream so you can rebase

> but when i try to build docker image for valgrind it fails in test/valgrind/Dockerfile:14 COPY . .
ERROR: failed to solve: cannot copy to non-directory: /var/lib/docker/overlay2/long-random-hash/merged/bin

I've never had that issue and I'm not Docker expert. Could you dig into it? And if you do, self-answer here for later users! I could try to help soonish but I have very limited bandwidth unfortunately...

balbesina:
You are right, after changing `work_dir=/app` docker built it just fine.<br>
In order to generate suppression i uncommented `valgrind_generate_suppressions: true` in the rakefile. And when I added it to supp file the initial error disappears. However, new memcheck error shows up and I don't see any connection between those methods (my makeValid changes and `method_geometry_contains` = which is an equivalent for "contains?" ruby method).

```
288 (24 direct, 264 indirect) bytes in 1 blocks are definitely lost in loss record 16,757 of 20,094
malloc (vg_replace_malloc.c:381)
objspace_xmalloc0 (gc.c:12617)
rb_id_table_create (id_table.c:98)
cache_callable_method_entry (vm_method.c:1439)
callable_method_entry_or_negative (vm_method.c:1497)
callable_method_entry (vm_method.c:1517)
rb_callable_method_entry (vm_method.c:1524)
gccct_method_search_slowpath (vm_eval.c:456)
gccct_method_search (vm_eval.c:505)
rb_call0 (vm_eval.c:546)
rb_class_new_instance_kw (object.c:2152)
rb_exc_new (error.c:1385)
*rbimpl_exc_new_cstr (string.h:1482)
*rgeo_convert_to_geos_geometry (factory.c:849)
*method_geometry_contains (geometry.c:605)
vm_call0_cfunc_with_frame (vm_eval.c:173)
vm_call0_cfunc (vm_eval.c:187)
vm_call0_body (vm_eval.c:233)
vm_call0_cc (vm_eval.c:110)
rb_vm_call0 (vm_eval.c:70)
rb_vm_call_kw (vm_eval.c:330)
vm_call_cfunc_with_frame_ (vm_insnhelper.c:3502)
vm_call_cfunc_array_argv (vm_insnhelper.c:3581)
vm_call_cfunc_only_splat (vm_insnhelper.c:3602)
vm_sendish (vm_insnhelper.c:5593)
vm_exec_core (insns.def:834)
rb_vm_exec (vm.c:2486)
vm_yield_with_cref (vm.c:1634)
vm_yield (vm.c:1642)
rb_yield_0 (vm_eval.c:1366)
rb_yield (vm_eval.c:1382)
rb_ary_each (array.c:2538)
vm_call_cfunc_with_frame_ (vm_insnhelper.c:3502)
vm_call_cfunc_with_frame (vm_insnhelper.c:3530)
vm_sendish.constprop.0 (vm_insnhelper.c:5593)
vm_exec_core (insns.def:814)
rb_vm_exec (vm.c:2486)
vm_yield_with_cref (vm.c:1634)
vm_yield (vm.c:1642)
rb_yield_0 (vm_eval.c:1366)
rb_yield (vm_eval.c:1382)
rb_ary_collect (array.c:3633)
vm_call_cfunc_with_frame_ (vm_insnhelper.c:3502)
vm_call_cfunc_with_frame (vm_insnhelper.c:3530)
vm_sendish.constprop.0 (vm_insnhelper.c:5593)
vm_exec_core (insns.def:814)
rb_vm_exec (vm.c:2486)
rb_proc_call_kw (proc.c:957)
exec_end_procs_chain (eval_jump.c:105)
rb_ec_exec_end_proc (eval_jump.c:120)
rb_ec_teardown (eval.c:159)
rb_ec_cleanup (eval.c:212)
ruby_run_node (eval.c:328)
rb_main (main.c:39)
main (main.c:58)
```

balbesina:
hey, @keithdoggett. tests were added. pls review

ext/geos_c_impl/geometry.c
```diff
@@ -1142,8 +1183,13 @@ method_geometry_make_valid(VALUE self)
   if (!self_geom)
     return Qnil;
 
+  GEOSMakeValidParams* params = GEOSMakeValidParams_create();
+  rgeo_make_valid_apply_params(argc, argv, params);
```
BuonOmo, line 1187: This method must not raise in order to keep our garbages properly collected. Or it can raise (`rb_funcall` for instance) or `rgeo_make_valid_check_method` which can raise. So in the current state of the code you have a potential leak. You have to secure this bit of code. In the current state I suggest using rb_protect directly.


See https://ulysse.md/2023/01/24/Calling-Ruby-Methods-in-C-Avoid-Memory-Leaks.html (`rb_protect`) or https://blog.peterzhu.ca/ruby-c-ext-part-8/ (`rb_ensure`)

test/valgrind/suppressions/ruby.supp
```diff
@@ -14,3 +14,12 @@
   fun:memmove
   fun:flo_to_s
 }
+{
+  <make_valid_false_positive>
```
BuonOmo, line 18: I don't think this is false positive (it never is tbh ^^)

BuonOmo (changes requested):
I think I found where your new suppression come from !

BuonOmo:
Actually I just saw that there is now a `RUBY_FREE_AT_EXIT` env variable that allow us to remove the suppression file ! (https://blog.peterzhu.ca/finding-memory-leaks-in-the-ruby-ecosystem/)

test/valgrind/suppressions/ruby.supp
keithdoggett: I merged in your PR with the upgraded memcheck functionality so this should be able to be removed now once `main` is merged in with this branch.

test/geos_capi/multi_polygon_test.rb
```diff
@@ -37,4 +37,55 @@ def test_polygonize
 
     assert_equal expected, input.polygonize
   end
+
+  INVALID_MULTIPOLYGON = "MULTIPOLYGON (((0 1,2 1,2 2,0 2,0 1)), ((1 0,3 0,3 3,1 3,1 0)))"
+  LINEWORK_MULTIPOLYGON = "MULTIPOLYGON (((1 2, 1 1, 0 1, 0 2, 1 2)), " \
+                          "((1 3, 3 3, 3 0, 1 0, 1 1, 2 1, 2 2, 1 2, 1 3)))"
+  JOINED_POLYGON = "POLYGON ((0 2, 1 2, 1 3, 3 3, 3 0, 1 0, 1 1, 0 1, 0 2))"
```
keithdoggett, line 44: one small nitpick, but instead of defining these as constants can we make these private methods in the test class that return the wkt. A similar thing was done for a lot of the validity checks if you want something to reference. Keeping it as WKT is fine (I believe most of the validity tests return geometry objects), but this will make the style more consistent.

keithdoggett: same with the line_string test as well please

ext/geos_c_impl/geometry.c
keithdoggett: Agree that we should secure this, but what would we need to free in this case? Seems like not much has been done by this point and a failure in `rgeo_make_valid_check_method` already frees the params.

keithdoggett (commented):
The tests are great thank you for handling that. I agree with @BuonOmo we should secure that one point to ensure there's no memory leak, but like I said in the comment, I think `rb_protect` should be easy enough since we shouldn't have to free too much that early in the function.

ext/geos_c_impl/geometry.c
BuonOmo: We need to free the params! We we call `scan_args` or `rb_funcall`, these are also subjects to a jump unfortunately (and I guess `rb_get_kwargs` could jump as well).

test/geos_capi/multi_polygon_test.rb
balbesina: that was done

ext/geos_c_impl/geometry.c
balbesina: i changed it to ruby kwarg processing to avoid c complexity. but not sure if it fits your high standards 🫣

BuonOmo: I'm sorry you're feeling it like this... This C code can be really tricky, and easily lead to memory leaks. I seems like the new way you are doing it should work, let's see what memcheck has to say. If that works that way I feel like we should make the `geometry_make_valid` private to make sure one is not calling it with wrong arguments. WDYT?

balbesina: sure, i've applied requested change. thx for review

BuonOmo (commented):
LGTM overall. I just have a tiny doubt on changing the `make_valid` method in `validity_check.rb`, which is supposed to be generic. Also we loose the check of `ValidityCheck` loading before any call to `make_valid` is made. I remember this being pretty important. I don't think I'll have time to check this in the next few days. Maybe you have an idea on that @keithdoggett. Otherwise if you prefer @balbesina we can roll back to the c version and just fix it (If you're ok I could do it, no big deal). What do you guys think ?

balbesina:
> Also we loose the check of ValidityCheck loading before any call to make_valid is made. I remember this being pretty important

could someone explain the importance of validity_check for `#make_valid` method to me? i see that this module was totally devoted to `#check_validity!`, but not for `#make_valid`.

> Otherwise if you prefer @balbesina we can roll back to the c version and just fix it

i was thinking about my initial implementation for all these weeks and it bothered me. c, cpp in geos work with 0/1 integers for makeValidWithParams. adding artificial string/symbol parameter `method` handling to c level feels bad. and since we have a mixture of ruby+c, then why not to split the responsibility?

but its up to you, maintainers, to make a decision.

BuonOmo:
> and since we have a mixture of ruby+c, then why not to split the responsibility?

The issue is not about splitting the responsibility here, I definitely like this idea actually. The issue is that this module is not supposed to be geos specific (other implementations are welcome!). And we make it geos specific here.

> could someone explain the importance of validity_check for #make_valid method to me? i see that this module was totally devoted to #check_validity!, but not for #make_valid.

That's actually a fair point. After a brief check I cannot see why we did it nor why we did it for `invalid_reason`. I think it is written somewhere in our PR history. I'll check further as I know this was a tedious implementation and there might be a very precise reason. @keithdoggett any memory about it?

keithdoggett:
@balbesina @BuonOmo the reason we included `make_valid` in that module is because this, in combination with `valid_op` is supposed to be a generic interface for validity testing and we figured lumping in `make_valid` with this makes sense.

I agree with @BuonOmo we should not modify the generic definition in `ValidityCheck`. Since this is just a C implementation, we can override `make_valid` in the `Geos` classes only (similar to the FFI classes) and have that handle the kwargs. I'd be open to modifying the generic implmentation to handle args in case they are accidentally passed to a cartesian factory geometry for example and we don't want an argument error to be raised (though an error would be raised anyways since we don't implement `make_valid` for them).

balbesina:
@BuonOmo I still believe there is an issue with valgrind that requires suppression. Or else, please, explain how to avoid "memory loss".

I did several tests with different ways of processing. Removed absolutely everything except for `GEOSMakeValidParams_create` and `GEOSMakeValidParams_destroy` calls: there is no kwargs parsing, there is no arity check, there are no extra variables that are defined. Only params for make valid.

in this case valgrind raises memory loss 
`8 bytes in 1 blocks are definitely lost in loss record 3 of 96`
```C
static VALUE
method_geometry_make_valid(int argc, VALUE* argv, VALUE self)
{
  RGeo_GeometryData* self_data;
  const GEOSGeometry* self_geom;
  GEOSGeometry* valid_geom;
  self_data = RGEO_GEOMETRY_DATA_PTR(self);
  self_geom = self_data->geom;
  if (!self_geom)
    return Qnil;

  GEOSMakeValidParams* params = GEOSMakeValidParams_create();
  valid_geom = GEOSMakeValid(self_geom);
  GEOSMakeValidParams_destroy(params);

  if (!valid_geom) {
    rb_raise(rb_eRGeoInvalidGeometry,
             "%" PRIsVALUE,
             method_geometry_invalid_reason(self));
  }
  return rgeo_wrap_geos_geometry(self_data->factory, valid_geom, Qnil);
}
```

in this case there is no valgrind error error
```C
static VALUE
method_geometry_make_valid(int argc, VALUE* argv, VALUE self)
{
  RGeo_GeometryData* self_data;
  const GEOSGeometry* self_geom;
  GEOSGeometry* valid_geom;
  self_data = RGEO_GEOMETRY_DATA_PTR(self);
  self_geom = self_data->geom;
  if (!self_geom)
    return Qnil;

  GEOSMakeValidParams* params = GEOSMakeValidParams_create();
  GEOSMakeValidParams_destroy(params);
  valid_geom = GEOSMakeValid(self_geom);

  if (!valid_geom) {
    rb_raise(rb_eRGeoInvalidGeometry,
             "%" PRIsVALUE,
             method_geometry_invalid_reason(self));
  }
  return rgeo_wrap_geos_geometry(self_data->factory, valid_geom, Qnil);
}
```

As you can see I don't even call `GEOSMakeValidWithParams`. Obviously my tests fail, but valgrind shows different results for those examples. Why?

BuonOmo:
@balbesina I'm quite late I know but I wanted to have some closure on the PR. Unfortunately I cannot push on your branch. Could you either give the rights, or allow that I swap to another PR?

balbesina:
@BuonOmo  added you as a collaborator on the fork, is it enough?

BuonOmo:
Yes! Thank you :)

FYI for now I only updated the branch to see what's failing now that memcheck evolved a bit since last time

BuonOmo:
I believe we're there. I would just like to understand what motivated your last commit ? I would never have supposed a non-ruby call would proc a jump within ruby. And I still do not understand it ^^'

Besides, LGTM!

@keithdoggett anything in mind?

Thank you for staying motivated in pushing this pr forward

balbesina:
globals.c:94 registers error handler that calls rb_raise. bowtie polygon make_valid runs it and throws UnsupportedOperationException from method GEOSMakeValidWithParams. since rb_raise does longjmp - c-ish destroy is not called. thanks for the review

PR #373, opened 2024-09-15
Adds distance method from spherical LineString to Point

eikes:
This PR implements two new methods to the `SphericalLineStringMethods` module, both can only take a `SphericalPointImpl` as a parameter:

`#distance`
`#closest_point`

The implementation requires the closest_point method to also be present in the internal `RGeo::Geographic::SphericalMath::ArcXYZ` class. This in turn depends on the ability to project a point on the unit sphere onto the great arc of a different arc. This is implemented in the `project_point` method, it introduces the building block which the other methods depend on. The other methods are more or less trivial.

### DIFF
--- a/lib/rgeo/geographic/spherical_feature_methods.rb
+++ b/lib/rgeo/geographic/spherical_feature_methods.rb
@@ -40,6 +40,8 @@ module RGeo
         case rhs
         when SphericalPointImpl
           xyz.dist_to_point(rhs.xyz) * SphericalMath::RADIUS
+        when SphericalLineStringImpl
+          rhs.distance(self)
         else
           super
         end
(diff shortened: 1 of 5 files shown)


keithdoggett (approved):
@eikes thank you for the PR. The tests are clear and the code looks good to me. Just one question. How hard would it be to implement linestring to linestring distance now? Seems like it should be an easy enough lift given that you've already done the hard part here. Not necessary to get this PR merged but if it's easy enough could make sense to do now too.

eis-ioki:
@keithdoggett I don't think the distance of two arcs on a sphere is ever anything other than the minimum of the distances of all enpoints to the other arc, unless they cross each other:

```ruby
def distance(arc_1, arc_2)
  return 0 if arc_1.crosses?(arc_2)
  [
    arc_1.s.distance(arc_2),
    arc_1.e.distance(arc_2),
    arc_2.s.distance(arc_1),
    arc_2.e.distance(arc_1)
  ].min
end
```

eikes:
@keithdoggett I don't think the distance of two arcs on a sphere is ever anything other than the minimum of the distances of all enpoints to the other arc, unless they cross each other:

```ruby
def distance(arc_1, arc_2)
  return 0 if arc_1.crosses?(arc_2)
  [
    arc_1.s.distance(arc_2),
    arc_1.e.distance(arc_2),
    arc_2.s.distance(arc_1),
    arc_2.e.distance(arc_1)
  ].min
end
```

eikes:
I am getting unexpected results here. I am reverting to draft status till I know why...

eikes:
@keithdoggett I have finally figured out why my real life test was failing. It now passes. 

The project_point method was return a point that, due to floating point math, was just a little bit off the plane. That in turn led to contains_point? returning false, which in the end meant that it didn't return a point on the arc, but rather an endpoint of the linestring.

BuonOmo:
> The project_point method was return a point that, due to floating point math, was just a little bit off the plane.

Interesting, it seems related to #395. We may have a global issue on handling floating point errors

PR #385, opened 2025-11-25
fix: Ruby 4.0 compatibility

seuros:
Replace deprecated typed_flag check with RTYPEDDATA_P/RTYPEDDATA_TYPE macros

was deprecated since ruby 3.


BuonOmo:
> sure i can work on reactor issue.

That was not a request, but feel free!

> Should i edit the commit msg, or you will do it when you squash/merge ?

I can, just waiting for a second review at this stage

> PS: the rubocop error, is just rubocop not installing in with the new bundler, it will get fixed upstream

Yes, I saw that, let's make it another PR

keithdoggett:
Looks like we need to modify the gemspec for the rubocop check to run

PR #386, opened 2025-12-15
chore: add Ruby 4.0 compatibility gems and expand CI matrix

seuros:
@BuonOmo @keithdoggett 

Here is the fix for a clean build

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,7 @@ jobs:
       matrix:
         ruby:
           - head
-          - "4.0.0-preview3"
+          - "4.0"
           - "3.4"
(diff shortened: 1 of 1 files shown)


BuonOmo (commented):
Thank you! Could we remove the 3.1 since it is EOL and maybe the 4.0.0-preview2 as head is already at 4.0.0.0-dev ?

seuros:
Let keep for now

seuros:
The preview3 will be replaced with 4.0.0-stable after Christmas.

the head will stay target 4.1 after that.

That allow us to see incoming API change.  You can edit the file and remove head if you wish.

BuonOmo (approved):

PR #388, opened 2026-01-08
fix: mark error_handler as NORETURN to silence compiler warning

seuros:
The error_handler function always calls rb_raise which never returns. 
Adding NORETURN attribute eliminates the -Wmissing-noreturn warning when compiling with Ruby 4.1-dev.


BuonOmo (approved):
I know that I did not do that on purpose for some reason, maybe macos dev. But I don't remember the precise reason so let's ship and I'll rollback if needed

PR #392, opened 2026-01-22
feat: support of 3/4D, DBSCAN...

seuros:
This PR might or might not be merged as-is or split

I cherry-picked commits from my actual usage.

The changes expose current libgeos APIs that provide better performance and leverage native functionality:
  - Native M-coordinate support (4D geometries)
  - DBSCAN clustering
  - Coverage operations
  - line_substring, simple_detail

The upgrade is backward compatible unless user is running an EOL distro. 
As discussed in the issue, any new API calls will fail at runtime if libgeos doesn't support them, that's a user responsibility, not the library's. Previous usage will be fine.

Future possibilities: curved geometries, ConcaveHull improvements, MaximumInscribedCircle, LargestEmptyCircle not needed for my use case but nice to have later. 

---

In other note: rgeo gem always get compiled in the host machine, that means user need to have the header. 

I could modify our release flow to ship precompiled gem like what nokogiri does, CI will not need to install the header.


BuonOmo:
Can you please make multiple PRs for various topics. I cannot afford to spend that much time on reviewing a massive PR, sorry.

Also, please try minimal and focus changes (I see for instance that you are removing method comments, that would need to be a focus change on its own, or at least explained).

And can you start by making PRs that DO NOT require a breaking change ? So we can ship the most possible without the need of backporting.

seuros:
Absolutely. As i stated  this was just a cherry pick from my local work (https://github.com/rgeo/rgeo/pull/392/commits) you can see those were done last year.  I also did not want to vanish for few weeks/months after we just closed the issue.

This PR is not to review the code, i will have to split it. But to gather feedback from you and others on what make sense to get merged.

For example : do we need DBSCAN clustering in core ?  Do others want features in Libgeos that i'm not aware of ?

If you see anything from the description that you find interesting , i'm willing top open a separate PR.

The removable will be reverted, i just did not want to spend time editing the commit.

I converted this as Draft to mark the indent.

Issue #393, opened 2026-03-17
RGeo 3.x throwing "Self Intersecting" errors on what should be valid, non-intersecting polygons


scott-lewis-nsidc:
# Steps to reproduce

Running the following code snippet (I called it `test_rgeo.rb`:

```
require 'rgeo'

factory = RGeo::Cartesian.simple_factory(:srid => 4326)
bbox = [-180.0, -85.044, 180.0, 90.0]
minx, miny, maxx, maxy = bbox

p1 = factory.point(minx, maxy)
p2 = factory.point(maxx, maxy)
p3 = factory.point(maxx, miny)
p4 = factory.point(minx, miny)

ring = factory.linear_ring([p1, p2, p3, p4, p1])
poly = factory.polygon(ring)

puts "IS RING VALID? #{ring.valid?}"
puts "IS POLY VALID? #{poly.valid?}"

puts "RING INVALID REASON: #{ring.invalid_reason}"
puts "POLY INVALID REASON: #{poly.invalid_reason}"

puts "POLY ENVELOPE: #{poly.envelope}"
```

# Expected behavior
```
IS RING VALID? true
IS POLY VALID? true
RING INVALID REASON: 
POLY INVALID REASON:
POLY ENVELOPE: POLYGON ((-180.0 -85.044, 180.0 -85.044, 180.0 90.0, -180.0 90.0, -180.0 -85.044))
```

# Actual behavior
```
IS RING VALID? false
IS POLY VALID? false
RING INVALID REASON: Self-intersection
POLY INVALID REASON: Self-intersection
/<PATH>/gems/ruby-3.2.9/gems/rgeo-3.1.0/lib/rgeo/impl_helper/validity_check.rb:105:in `check_validity!': Self-intersection (RGeo::Error::InvalidGeometry)
        from /<PATH>/gems/ruby-3.2.9/gems/rgeo-3.1.0/lib/rgeo/impl_helper/validity_check.rb:78:in `block (3 levels) in override'
        from test_rgeo.rb:21:in `<main>'
```

# System configuration
Ruby version: 3.2.9 (but I've experienced this in 3.2.2 as well)

OS: Pop!_OS 22.04 LTS

rgeo version: 3.1.0 (but also get this error with rgeo 3.0.0)

# Other notes
[...]

PR #394, opened 2026-03-19
Precision option for simple_factory

scott-lewis-nsidc:
### Summary

Adding a precision option to simple_factory to reduce incorrect self-intersection errors.

Addresses the issue I reported in https://github.com/rgeo/rgeo/issues/393

### Other Information

I was only able to reliably reproduce the bug with `simple_factory`, so I only added the precision option to that factory.  It could be added to other factories if needed.  I decided to go with a precision of 5 decimal places as this is still very accurate for most needs (and could be expanded if finer precision is needed).

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,10 @@
 ### Unreleased
 
+**Bug Fixes**
+
+* Add a `:precision` option for `simple_factory` instances to reduce invalid self-intersection issues
+* Add a new tests to validate `simple_factory` polygons with non-integer vertices
+
 ### 3.1.0 / 2025-01-20
 
 **Minor Changes**
--- a/lib/rgeo/cartesian/factory.rb
+++ b/lib/rgeo/cartesian/factory.rb
@@ -38,6 +38,8 @@ module RGeo
         @coord_sys = coord_sys_info[:coord_sys]
         @srid = coord_sys_info[:srid]
 
+        @precision = opts[:precision] || 5
+
         @buffer_resolution = opts[:buffer_resolution].to_i
         @buffer_resolution = 1 if @buffer_resolution < 1
 
@@ -186,7 +188,7 @@ module RGeo
       # See RGeo::Feature::Factory#point
 
       def point(x, y, *extra)
-        PointImpl.new(self, x, y, *extra)
+        PointImpl.new(self, x.round(@precision), y.round(@precision), *extra)
       end
 
       # See RGeo::Feature::Factory#line_string
--- a/test/common/validity_tests.rb
+++ b/test/common/validity_tests.rb
@@ -81,6 +81,7 @@ module RGeo
 
         def test_validity_invalid_reason
           assert_nil(square_polygon.invalid_reason)
+          assert_nil(square_polygon_decimals.invalid_reason)
           assert_equal("Self-intersection", bowtie_polygon.invalid_reason)
         end
 
@@ -115,6 +116,20 @@ module RGeo
                 @factory.point(1, 1),
                 @factory.point(0, 1),
                 @factory.point(0, 0)
+              ]
+            )
+          )
+        end
+
+        def square_polygon_decimals
+          @square_polygon_decimals ||= @factory.polygon(
+            @factory.linear_ring(
+              [
+                @factory.point(-180.0, 90.0),
+                @factory.point(180.0, 90.0),
+                @factory.point(180.0, -85.044),
+                @factory.point(-180.0, -85.044),
+                @factory.point(-180.0, 90.0)
               ]
             )
           )
--- a/lib/rgeo/cartesian/interface.rb
+++ b/lib/rgeo/cartesian/interface.rb
@@ -70,6 +70,11 @@ module RGeo
       # [<tt>:coord_sys_class</tt>]
       #   CoordSys::CS::CoordinateSystem implementation used to instansiate
       #   a coord_sys based on the :srid given.
+      # [<tt>:precision</tt>]
+      #   Set the rounding precision for new points created by this factory. This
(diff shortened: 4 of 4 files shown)


test/common/validity_tests.rb
```diff
@@ -120,6 +121,20 @@ def square_polygon
           )
         end
 
+        def square_polygon_decimals
```
scott-lewis-nsidc, line 124: These specific coordinates were chosen for the test as they were the ones I was getting failures for.  Oddly enough, though, the erroneous Self-Intersection was at the upper-left corner (near -180.0, 90.0), but for some reason doing the bounds of (-180.0, -90.0) to (180.0, 90.0) didn't produce the error; I'm not sure why changing the south boundary caused the issue to present itself.

lib/rgeo/cartesian/factory.rb
```diff
@@ -38,6 +38,8 @@ def initialize(opts = {})
         @coord_sys = coord_sys_info[:coord_sys]
         @srid = coord_sys_info[:srid]
 
+        @precision = opts[:precision] || 5
```
scott-lewis-nsidc, line 41: A precision of 5 decimal places is equivalent to about 1.11m in latitude/longitude coordinates.  That is probably plenty for the most common usages.

keithdoggett: This is probably reasonable, but this could be used for an alternative coordinate system so the range may not just be that of latitudes and longitudes. That said most other systems I've used have had larger ranges so this shouldn't be an issue.

keithdoggett (commented):
Hey @scott-lewis-nsidc thanks for working on this. I have a few notes. I still need to think this over some.

1. Regarding your original issue, the reason this has not come up before is that the `simple_factory` is probably not used a ton, likely only for simple geometric calculations. For cartesian operations most users will opt for using the GEOS library (see: https://github.com/rgeo/rgeo/blob/main/doc/Installing-GEOS.md) and for geographic work users will use the `geographic_factory`. You also got this error when upgrading because I reworked some of the internals to the ruby based cartesian implementation to support more operations and guess the intersection detection is prone to this kind of error.
2. The piece I'm considering more is what should the default be or should we have it not set by default (and no rounding is done) and it's only evaluated when someone sets the precision. We try to follow geos somewhat and their precision model defaults to floating point (https://libgeos.org/doxygen/classgeos_1_1geom_1_1PrecisionModel.html) but you can set a grid size. Now I like the idea of rounding better than creating an entire grid and pinning implementation here but I do worry that setting it by default at a relatively low precision is not the best idea.

Overall I think your implementation looks good just want to consider the default value and if we should really be setting it. But practically I think 5 is reasonable.

lib/rgeo/cartesian/factory.rb
scott-lewis-nsidc: Agreed.  For the coordinate systems we've used, lat/lon is by far the one with the "smallest" range in terms of integers.  Many of the ones we use have ranges in the millions, which would make the 5 decimal places even more minute.

In any case, I figured that's why having it be an option one could pass in would be helpful, in case someone needs a smaller coordinate system with finer resolution.

scott-lewis-nsidc:
> Hey @scott-lewis-nsidc thanks for working on this. I have a few notes. I still need to think this over some.
> 
>     1. Regarding your original issue, the reason this has not come up before is that the `simple_factory` is probably not used a ton, likely only for simple geometric calculations. For cartesian operations most users will opt for using the GEOS library (see: https://github.com/rgeo/rgeo/blob/main/doc/Installing-GEOS.md) and for geographic work users will use the `geographic_factory`. You also got this error when upgrading because I reworked some of the internals to the ruby based cartesian implementation to support more operations and guess the intersection detection is prone to this kind of error.

In our particular use case, we aren't really even doing much fancy with it - it's mostly used as a bounding box for a geographic region.  That's why it was stumping me for a bit as to why the self-intersection issue was happening, since it definitely wasn't anything fancy.  Using a geographic factory was a bit of overkill for what we were needing.

>     2. The piece I'm considering more is what should the default be or should we have it not set by default (and no rounding is done) and it's only evaluated when someone sets the precision. We try to follow geos somewhat and their precision model defaults to floating point (https://libgeos.org/doxygen/classgeos_1_1geom_1_1PrecisionModel.html) but you can set a grid size. Now I like the idea of rounding better than creating an entire grid and pinning implementation here but I do worry that setting it by default at a relatively low precision is not the best idea.
> 
> 
> Overall I think your implementation looks good just want to consider the default value and if we should really be setting it. But practically I think 5 is reasonable.

If needed, I could refactor the code an have no default, and then only round it if a precision is actually provided.  I did it this way first so that we could just plug in the new library and not have to go through our code to manually set the precision when instantiating the factory.

That said, especially if simple_factory is not used as often, I was thinking that a 5 decimal default would be a good compromise for most use cases.

In any case, I appreciate the consideration thus far!  Just let me know if you want me to change it.

scott-lewis-nsidc:
Sorry to "bump" this, but I thought I'd follow up and see if you've had a chance to give this some further consideration.

keithdoggett (approved):
@scott-lewis-nsidc sorry this slipped. I think this is fine especially since it can be set. I'll see if @BuonOmo has any thoughts but if not can get this merged soon.

History.md
```diff
@@ -1,5 +1,10 @@
 ### Unreleased
 
+**Bug Fixes**
+
+* Add a `:precision` option for `simple_factory` instances to reduce invalid self-intersection issues
```
BuonOmo, line 5: @keithdoggett sorry waking up super late. Shouldn't this be a minor change ? Even maybe major. It does affect the API. Or we do not default to 5 for the precision ?

